PR #16, opened 2021-11-04
Add virtual walls and no mopping zones

edenhaus:


deebot_client/commands/map.py
```diff
@@ -111,7 +123,7 @@ def _handle_body_data_dict(
         """
         args = {
             cls._ARGS_ID: data["mid"],
-            cls._ARGS_SET_ID: data["msid"],
+            cls._ARGS_SET_ID: data.get("msid", None),
```
edenhaus, line 126: Only the change of this line is required to fix https://github.com/DeebotUniverse/Deebot-4-Home-Assistant/issues/48

PR #250, opened 2023-05-13
ignore errors in payload

edenhaus:
Fixes issue https://github.com/DeebotUniverse/Deebot-4-Home-Assistant/issues/344


edenhaus:
Fixed in #253

Issue #279, opened 2023-07-16
Map renderer crash due to off by one error


bitnimble:
### Checks

- [X] I have searched the existing issues and no issue is describing my issue
- [X] I have checked the [FAQ](https://deebot.readthedocs.io/home/faq/)
- [X] I have checked the [documentation](https://deebot.readthedocs.io/integrations/home-assistant)
- [X] I have installed the latest version

### The problem

`map.py` has an off-by-one error when rendering dashed lines that end up causing a divide by zero: https://github.com/DeebotUniverse/client.py/blob/ebfae315f72196626ba4e05d37d317bc46aceb23/deebot_client/map.py#L505. I guess this is just a rare case when somebody's map specifically has two points to be rendered that are on top of each other. 

When length is 0, it'll still perform the loop and crash. 

Updated it to `while position < length:` locally and it now works perfectly on my map.

### On which deebot vacuum you have the issue?

Deebot T9

### Which version of the deebot-client are you using?

2.0.1

### Country

ww

### Continent

au

edenhaus:
If you would like you can directly create a PR for this fix

bitnimble:
yep cool will do, just wanted to double check first in case you might already have a fix, or if there's something else about this code that i'm not aware of 🙂

PR #280, opened 2023-07-17
map: fix divide-by-zero in dashed_line renderer

bitnimble:
Fixes https://github.com/DeebotUniverse/client.py/issues/279


edenhaus:
Thanks @bitnimble :)

PR #348, opened 2023-11-15
Add "Ota" commands

MVladislav:
# Add new command: "Ota"

> See more general info and details under **[discussion](https://github.com/DeebotUniverse/client.py/discussions/345)**.

- Tested with **Deebot T10** over HA - [Deebot-4-Home-Assistant](https://github.com/MVladislav/Deebot-4-Home-Assistant/tree/p95mgv) integration
- Commands
  - `getOta`
  - `setOta`

| Ecovacs Home App  | Home Assistant |
| :---------------- | :------------- |
| ![setOta][setOta] |   ![setOta_home][setOta_home] |

[setOta]: https://github.com/DeebotUniverse/client.py/assets/7400017/ca6ab521-30fe-4336-a5e8-1ada8d312480
[setOta_home]: https://github.com/DeebotUniverse/client.py/assets/7400017/26881fc8-fc7d-4e04-9f7c-39c7441fd39b


deebot_client/commands/json/ota.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @classmethod
+    def _handle_body_data_dict(
+        cls, event_bus: EventBus, data: dict[str, Any]
+    ) -> HandlingResult:
+        """Handle message->body->data and notify the correct event subscribers.
+
+        :return: A message response
+        """
+        event_bus.notify(cls.event_type(bool(data["autoSwitch"])))
```
edenhaus, line 27: The deebot 950 Ozmo will not return the `autoSwitch` property but other data.
We should not expect that `autoSwitch` needs to be always present.

What is your bot returning on this command?

MVladislav: Is it possible to turn off **auto ota** for **deebot 950 Ozmo**? 
I do not find the option for **950** in the APP and the result over API has also not that option (like you say **autoSwitch** not available).

The **deebot t10** has the option in the APP and over API. 
From what i see, this option is only available for newer bot's.

**t10** returns (https://github.com/DeebotUniverse/client.py/discussions/345#discussioncomment-7582167):
image

**950** return:
image

edenhaus: The 950 has no auto-update. You can only update it manually

edenhaus: I think `supportAuto` indicates if auto-update is available or not.
We should adapt the `OtaEvent` to have at least the current version, the status and the progress.

Not sure what's the meaning of result and isForce

MVladislav: What do you think about the updated version:

- extended with additional information for **get**
  - added: `support_auto`, `version`, `status`, `progress`
  - on inital **get** it is possible all the values could be filled
    - for older bots `support_auto` is set to `None` when not send and `enable` will by default `True`
    - for newer bots `support_auto` and `enable` will be converted as bool
- when **set** will be called to change `enable` state, only `autoSwitch` is respond by api
  - for that, all extended additional information, are set as None to identify no changes respond
  - for older bots, which can not change `autoSwitch`, they will have only **get** event and the state is always `enable=True`
- I'm not sure for older bots if they are auto updated or it is needed to manually trigger it in the app
  - Not saw updates in the app for long time
  - Here would be the question if `enable=True` is the correct state to show
- About the extended information `status` and `progress`
  - They will be only set by **get**
  - I'm not sure if there is an event when updates are happen

deebot_client/commands/json/ota.py
```diff
@@ -0,0 +1,51 @@
...
+class SetOta(JsonSetCommand):
+    """Set ota command."""
+
+    name = "setOta"
+    get_command = GetOta
+
+    _mqtt_params = MappingProxyType({"autoSwitch": InitParam(bool, "enable")})
+
+    def __init__(self, enable: bool) -> None:  # noqa: FBT001
+        super().__init__({"autoSwitch": 1 if enable else 0})
```
edenhaus, line 51:
```suggestion
    _mqtt_params = MappingProxyType({"autoSwitch": InitParam(bool, "auto_enabled")})

    def __init__(self, auto_enabled: bool) -> None:  # noqa: FBT001
        super().__init__({"autoSwitch": 1 if auto_enabled else 0})
```

edenhaus (changes requested):
Does there exists also a `onOta` message/event?

deebot_client/commands/json/ota.py
```diff
@@ -0,0 +1,51 @@
...
+        cls, event_bus: EventBus, data: dict[str, Any]
+    ) -> HandlingResult:
+        """Handle message->body->data and notify the correct event subscribers.
+
+        :return: A message response
+        """
+        support_auto = data.get("supportAuto")
+        ota_event = OtaEvent(
+            enable=bool(data.get("autoSwitch", True)),
+            support_auto=None if support_auto is None else bool(support_auto),
```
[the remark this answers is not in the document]
MVladislav: > Does there exists also a onOta message/event?

no, i saw only get and set

> Older bots must be updated manually

when the process for ota would be:
- on first init **get** will trigger and will informe with all possible values
- when **set** is used only `auto_enabled` is returned

with current version it would be:
```sh
# on init get
OtaEvent(enable=True, support_auto=True, version='1.7.2', status='idle', progress=0)

# when switch with set
OtaEvent(enable=False, support_auto=None, version=None, status=None, progress=None)
OtaEvent(enable=True, support_auto=None, version=None, status=None, progress=None)
OtaEvent(enable=False, support_auto=None, version=None, status=None, progress=None)
OtaEvent(enable=True, support_auto=None, version=None, status=None, progress=None)
```

with your change it would be:
```sh
# on init get
OtaEvent(enable=True, support_auto=True, version='1.7.2', status='idle', progress=0)

# when switch with set
OtaEvent(enable=False, support_auto=False, version=None, status=None, progress=None)
OtaEvent(enable=True, support_auto=False, version=None, status=None, progress=None)
OtaEvent(enable=False, support_auto=False, version=None, status=None, progress=None)
OtaEvent(enable=True, support_auto=False, version=None, status=None, progress=None)
```

when this information would be used, `support_auto` would be set with a wrong state as `False`, thats why i checked if `support_auto` is provided, and if not set to `None` to identify no change provided.

MVladislav:
> We need here also to split `OnOta` and `GetOta` if we are expecting events on `atr`

saw only get and set

edenhaus (changes requested):

deebot_client/commands/json/ota.py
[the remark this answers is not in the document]
edenhaus: This your concern solved with the latest implementation?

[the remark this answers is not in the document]
MVladislav: yes, looks good so far with the latest update

edenhaus (approved):

PR #352, opened 2023-11-16
Add new command: "ClearMap"

MVladislav:
# Add new command: "ClearMap"

- Tested with **Deebot 950** over HA - [Deebot-4-Home-Assistant](https://github.com/MVladislav/Deebot-4-Home-Assistant/tree/clear_map) integration
- Commands
  - `clearMap`
- Notes
  - ClearMap would be also possible for other bot's, but here is more research needed.
  - For **Deebot 950** the map can be cleared by only send `{"type": "all"}`.
  - Other bot's like **Deebot T10** needs for example `{"act": "del", "mid": "<REPLACED_MID>"}`, where i current do not know, how to collect the **mid** number.

| Ecovacs Home App  | Home Assistant |
| :---------------- | :------------- |
| ![clearMap][clearMap] |   ![clearMap_home][clearMap_home] |

[clearMap]: https://github.com/DeebotUniverse/client.py/assets/7400017/d6764145-d498-4861-914e-5caff05172bb
[clearMap_home]: https://github.com/DeebotUniverse/client.py/assets/7400017/dfe131bd-4586-4e4a-ab41-00314c7ea045


edenhaus:
Not sure if we should add this command to HA as all the other Map modifications commands are missing...
A user could accidentally delete the map but to save it again, he needs the app...

MVladislav:
> Not sure if we should add this command to HA as all the other Map modifications commands are missing... A user could accidentally delete the map but to save it again, he needs the app...

I add it because i need it for my 950, as the map is sometimes broken/duplicated and i need to reset it always over the app. 
For this bot i not use any special map function and only use to auto map generation, so only reset it and next clean it will auto generated.
To avoid in HA accidentally delete, the button can be disabled.

PR #350, opened 2023-11-15
Add new command: "clean log v2"

MVladislav:
# Add new command: "clean log v2"

> See more general info and details under **[discussion](https://github.com/DeebotUniverse/client.py/discussions/345)**.

- Tested with **Deebot T10** over HA - [Deebot-4-Home-Assistant](https://github.com/MVladislav/Deebot-4-Home-Assistant/tree/p95mgv) integration
- Info
  - added api for `app/dln/api/log/clean_result/list`
  - authentication is updated for use of different api prefix
  - possible this will not fully work with official api, as the request send by POST and should be a GET, i tested it with the offline bumper version

| Ecovacs Home App              | Home Assistant                                                                      |
| :---------------------------- | :---------------------------------------------------------------------------------- |
| ![clean_log_v2][clean_log_v2] | ![clean_log_v2_home][clean_log_v2_home] ![clean_log_v2_home_2][clean_log_v2_home_2] |

[clean_log_v2_home_2]: https://github.com/DeebotUniverse/client.py/assets/7400017/ef1b4c1f-8cee-4f5d-973d-e6133c4523fd
[clean_log_v2_home]: https://github.com/DeebotUniverse/client.py/assets/7400017/94cb4102-a665-4daf-986f-96df01a5949b
[clean_log_v2]: https://github.com/DeebotUniverse/client.py/assets/7400017/c9b8a5b5-0a47-40ae-a666-106df48bcded

PR #353, opened 2023-11-16
Add capabilities for Deebot T10 Plus (p95mgv)

MVladislav:
## Add bot definition: DEEBOT T10 PLUS :: p95mgv

> See more general info and details under **[discussion](https://github.com/DeebotUniverse/client.py/discussions/345)**.

### DIFF
--- a/deebot_client/capabilities.py
+++ b/deebot_client/capabilities.py
@@ -122,7 +122,7 @@ class CapabilityClean:
     action: CapabilityCleanAction
     continuous: CapabilitySetEnable[ContinuousCleaningEvent]
     count: CapabilitySet[CleanCountEvent, int] | None = None
-    log: CapabilityEvent[CleanLogEvent]
+    log: CapabilityEvent[CleanLogEvent] | None = None
     preference: CapabilitySetEnable[CleanPreferenceEvent] | None = None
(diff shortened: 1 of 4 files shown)


deebot_client/capabilities.py
```diff
@@ -122,7 +122,7 @@ class CapabilityClean:
     action: CapabilityCleanAction
     continuous: CapabilitySetEnable[ContinuousCleaningEvent]
     count: CapabilitySet[CleanCountEvent, int] | None = None
-    log: CapabilityEvent[CleanLogEvent]
+    log: CapabilityEvent[CleanLogEvent] | None = None
```
edenhaus, line 125: Why must these be optional?

MVladislav: Because the T10 is not use the current implemented log version, for what the issue #350 is for. 
In the p95mgv.py definition I not include the log part.

edenhaus: Please revert this change as otherwise the HA component needs an update too and we merge it after #350 is merged

edenhaus: We can leave it

edenhaus (approved):

PR #364, opened 2023-12-08
Add capabilities for T20e

marktdeng:
Added a symlink to t20 omni. It is the same except with cold water rather than hot water mopping.


edenhaus:
Thanks for your contribution :) I only have a quick question:

> It is the same except with cold water rather than hot water mopping.

Is this a hardware limitation, or can you set this somewhere in the UI?

marktdeng:
> Thanks for your contribution :) I only have a quick question:
> 
> > It is the same except with cold water rather than hot water mopping.
> 
> Is this a hardware limitation, or can you set this somewhere in the UI?

I only have the T20e so I'm not entirely sure, but I believe it's a purely hardware limitation.

bakerboy908:
@marktdeng Does the work mode selection appear/work for your t20e?

PR #288, opened 2023-08-06
Add support for MQTT + XML based devices

sandervankasteel:
This PR is for laying the ground work for supporting MQTT + XML devices. I've tested this against my own Deebot 900.

~Please note, this PR is still very much a WIP.~

**At least** the following things need to be done:
- [x] Completely implement the following commands
	- [x] Clean
		- [x] Sending
		- [x] Response
	- [x] Charge
		- [x] Sending
		- [x] Response
	- [x] CleanLogs (Completely adheres to JSON-API version :smile:)
		- [x] Sending
		- [x] Response
	- [x] ChargeState
		- [x] Sending
		- [x] Response
			- ~Currently will always report back as being docked~
	- [x] BatteryInfo
		- [x] Sending
		- [x] Response
	- [X] LifeSpan
			 This has been split up in 3 different commands (`GetLifeSpanBrush`, `GetLifeSpanSideBrush`, `GetLifeSpanHeap`) because the XML based devices does not have 1 endpoint for getting all the statistics on all the consumables. 
	- [x] FanSpeed (a.k.a CleanSpeed for the MQTT based devices)
		- [x] Getting
			- [x] Sending
			- [x] Response 
		- [x] Setting 
			- [x] Sending
			- [x] Response
	- [x] GetError
		- Has been implemented, but might benefit from some scenario testing. Since by default it appears that the `GetError` returns an empty string in the case of no errors.
- [x] Implement functionality for throwing exceptions for potentially unsupported commands on the older MQTT + XML devices
- [x] Add automated testing
- [x] Error responses
	Apparently the XML based devices use the same error codes as the JSON based devices.

~But with this PR, I'd like to ask for a preliminary code review. 
Just to get some initial feedback and see if I am on the correct track to get this merged.~

PR #369, opened 2023-12-12
Add xml GetError

edenhaus:
@sandervankasteel To go on with adding the xml commands, I think the best would be to split the PR #288 into smaller pieces. I have extracted the `GetError` command into this PR. Could you please test it?

### DIFF
--- a/deebot_client/commands/json/error.py
+++ b/deebot_client/commands/json/error.py
@@ -1,6 +1,7 @@
 """Error commands."""
 from typing import Any
 
+from deebot_client.const import ERROR_CODES
 from deebot_client.event_bus import EventBus
 from deebot_client.events import ErrorEvent, StateEvent
 from deebot_client.message import HandlingResult, MessageBodyDataDict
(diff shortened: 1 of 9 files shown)


lukakama:
@edenhaus Just tested the PR successfully on both newer T20 Omni (json) and older Ozmo 900 (xml). For the latter I had to define a device class named "y79a7u" with no capabilities except the error one, but it worked. 
For both i checked the "no errors", "off the floor" and "removed bin" statuses. They should be enough for the T20 Omni to assert no regressions.
For more advanced tests on the Ozmo 900 I need more time. Let me know if they are needed.

edenhaus:
@lukakama Thanks for testing and you need only to test this command on the 900 as it is a XML command. It should never be used for a json model like the new bots.

Can you please share me the capabilities file of the 900, so I will add it directly with this PR :)

lukakama:
For T20 I tested the JSON version of GetError command, as I noticed that it has been impacted by refactors needed to support the XML version (the externalization of error constants). It was just to assert absence of (imho very unlikely) regressions :P .

The capability file instead is just a sort of placeholder to perform a local test, which also does not honor type hints... Anyway, here it is:
> y79a7u.py

```py
"""Deebot Ozmo 900 Capabilities."""

from deebot_client.capabilities import (
    Capabilities,
    CapabilityEvent,
)
from deebot_client.const import DataType
from deebot_client.events import (
    ErrorEvent,
)
from deebot_client.models import StaticDeviceInfo
from deebot_client.util import short_name

from . import DEVICES

DEVICES[short_name(__name__)] = StaticDeviceInfo(
    DataType.XML,
    Capabilities(
        availability=None,
        battery=None,
        charge=None,
        clean=None,
        custom=None,
        error=CapabilityEvent(ErrorEvent, [GetError()]),
        fan_speed=None,
        life_span=None,
        map=None,
        network=None,
        play_sound=None,
        settings=None,
        state=None,
        stats=None,
        water=None,
    ),
)
```

PR #372, opened 2023-12-12
Add `GetMapSetV2` command

MVladislav:
#### Updates differences:

- **`GetCachedMapInfo`** is using **`GetMapSet`** to draw **"virtual walls"** or **"no map zones"**
  - for get it to work newer models use new API call over **`GetMapSetV2`** which can be choose by  **`GetCachedMapInfo`** on init by set **version = `1` or `2`**
- drawing the floor has an update, as the new bot's have more `pixel_type` which needs to be colored
  - over app they can be colored in different colors or also only in one color (here now only update for use one color)
  - **3** new types are added for coloring:
    - `0x04` => floor which could be related to a room but could not be reached by bot
    - `0x05` => any devices existing in a room like a couch, desk, ...
    - and every number not in `0-5`, is now a room, which could make it possible to color each room in a different color, but is here now color each room into a single color
    - [ ] **possible better colors to be choose**

#### Problems to be fixed:

- [ ] to be done is fix the event call for `GetMapSetV2`
  - the direct call of the function `GetMapSetV2` is working
  - but if the event is triggered by bot or app, it still use the old event and calls `GetMapSet` instead the new version
- [ ] test case not working
  - https://github.com/DeebotUniverse/client.py/blob/9f263b55b9efc55788f11f98c487af00c2a7a077/tests/commands/json/test_map.py#L119

#### Related issue:

- https://github.com/DeebotUniverse/Deebot-4-Home-Assistant/issues/268


deebot_client/commands/json/map.py
```diff
@@ -22,6 +24,14 @@ class GetCachedMapInfo(JsonCommandWithMessageHandling, MessageBodyDataDict):
     """Get cached map info command."""
 
     name = "getCachedMapInfo"
+    # version definition for using type of getMapSet v1 or v2
+    get_map_set_version: int = 1
```
edenhaus, line 28:
```suggestion
```

We shouldn't create a class variable for it

deebot_client/commands/json/map.py
```diff
@@ -22,6 +24,14 @@ class GetCachedMapInfo(JsonCommandWithMessageHandling, MessageBodyDataDict):
...
+
+    def __init__(
+        self, args: dict[str, Any] | list[Any] | None = None, version: int = 1
+    ) -> None:
+        self.get_map_set_version = version
```
edenhaus, line 33:
```suggestion
        match (version):
            case 1:
                self._map_set_command = GetMapSet
            case 2:
                self._map_set_command = GetMapSetV2
            case _:
                raise ValueError(f"version={version} is not supported")
```

We can directly store the type of the get command

deebot_client/commands/json/map.py
```diff
@@ -55,48 +65,15 @@ def _handle_response(
             return CommandResult(
                 result.state,
                 result.args,
-                [GetMapSet(result.args["map_id"], entry) for entry in MapSetType],
+                # an if check for getmapset, as newer bot use different api version
+                [GetMapSetV2(result.args["map_id"], entry) for entry in MapSetType]
+                if self.get_map_set_version == 2
+                else [GetMapSet(result.args["map_id"], entry) for entry in MapSetType],
```
edenhaus, line 71:
```suggestion
                [self._map_set_command(result.args["map_id"], entry) for entry in MapSetType],
```

I did not test this suggested changes, but imo they should work

deebot_client/commands/json/map.py
```diff
@@ -201,7 +178,7 @@ def __init__(
             type = type.value
 
         if msid is None and type == MapSetType.ROOMS.value:
-            raise ValueError("msid is required when type='vw'")
+            raise ValueError("msid is required when type='ar'")
```
edenhaus, line 181:
```suggestion
            raise ValueError(f"msid is required when type='{MapSetType.ROOMS.value}'")
```

deebot_client/commands/json/map.py
```diff
@@ -228,11 +205,19 @@ def _handle_body_data_dict(
...
                 name = cls._ROOM_NUM_TO_NAME.get(subtype, None)
 
+            coordinates: str | None = None
+            if data.get("compress", 0) == 1:
+                # NOTE: newer bot's return coordinates as base64 decoded string
+                coordinates = decompress_7z_base64_data(data["value"]).decode()
+            if coordinates is None:
+                # NOTE: older bot's return coordinates direct as comma separated list
+                coordinates = data["value"]
+
```
edenhaus, line 215: Is this change required when you created `GetMapSetV2`?

deebot_client/util.py
```diff
@@ -18,6 +24,28 @@ def md5(text: str) -> str:
     return hashlib.md5(bytes(str(text), "utf8")).hexdigest()  # noqa: S324
 
 
+def decompress_7z_base64_data(data: str) -> bytes:
+    """Decomporess base64 decoded 7z compressed string."""
+    _LOGGER.debug("[decompress7zBase64Data] Begin")
```
edenhaus, line 29:
```suggestion
```

A old left over and this log entry is a little bit useless

deebot_client/util.py
```diff
@@ -18,6 +24,28 @@ def md5(text: str) -> str:
...
+    for idx in decoded:
+        if i == 8:
+            final_array += b"\x00\x00\x00\x00"
+        final_array.append(idx)
+        i += 1
+
+    dec = lzma.LZMADecompressor(lzma.FORMAT_AUTO, None, None)
+    decompressed_data = dec.decompress(final_array)
+
+    _LOGGER.debug("[decompress7zBase64Data] Done")
```
edenhaus, line 45:
```suggestion
```

deebot_client/map.py
```diff
@@ -234,8 +229,13 @@ def _draw_map_pieces(self, draw: ImageDraw.ImageDraw) -> None:
                                 point_y,
                             )
                             raise MapError("Map Limit reached!")
-                        if pixel_type in [0x01, 0x02, 0x03]:
+                        if pixel_type in _COLOR_LIST_INCLUDE:
                             draw.point((point_x, point_y), fill=_COLORS[pixel_type])
+                        elif (
+                            pixel_type not in _COLOR_LIST_INCLUDE
+                            and pixel_type not in _COLOR_LIST_EXCLUDE
+                        ):
```
edenhaus, line 237:
```suggestion
                        elif pixel_type not in _COLOR_LIST_EXCLUDE:
```

The first statement can be omitted as it is always True.

Why do we need to draw all remaining values as floor?

deebot_client/commands/json/map.py
```diff
@@ -290,6 +326,42 @@ def _handle_response(
         return result
 
 
+class GetMajorMap(JsonCommandWithMessageHandling, MessageBodyDataDict):
```
edenhaus, line 329: Could you let me know why you moved this class? They are sorted alphabetically

edenhaus (changes requested):

deebot_client/commands/json/map.py
MVladislav: updated to use the match-case option

deebot_client/commands/json/map.py
MVladislav: updated to use the match-case option

deebot_client/commands/json/map.py
MVladislav: updated to use the match-case option

deebot_client/commands/json/map.py
MVladislav: done

deebot_client/commands/json/map.py
MVladislav: It has been a while since then, and I'm not entirely certain if I saw both possibilities at that point. This is why I decided to verify if compression was included, because i think i saw it is possible that some are compressed and some not. 

If you require a recheck, I can conduct some tests in the next few days.

deebot_client/util.py
MVladislav: removed

deebot_client/util.py
MVladislav: removed

deebot_client/map.py
MVladislav: removed in #373

deebot_client/commands/json/map.py
MVladislav: I have move it because of logical connection between the function as i tried to understand how the map draw is working. but in #373 it was moved back to sorted order.

MVladislav:
This is an image snippet how it looks with the SVG, color and map v2 changes with a T10 bot:
|current dev version|with map v2 command|
|:-|:-|
|image|image|

deebot_client/commands/json/map.py
edenhaus: If you have time to recheck that would be nice :)
As far as I know only V2 should support compression

deebot_client/commands/json/map.py
edenhaus: Moved it back to have a smaller changeset in this PR.
I think soon or later we need to create a module out of map

edenhaus:
I have made some minor improvements. I hope it is fine to push then directly on the PR



> This is an image snippet how it looks with the SVG, color and map v2 changes with a T10 bot:
> current dev version 	with map v2 command
> image 	image

Why are the rectangles filled with black?

deebot_client/commands/json/map.py
MVladislav: Removed.

I quick checked it and i think there should be no situation where this is needed,
as `MapSubsetEvent` is for `v2` notified in `GetMapSetV2`.
Possible i had problems at the beginnings because of the error described below, 
with remove of `v2` inside `deebot_client/messages/__init__.py`.

MVladislav:
> I have made some minor improvements. I hope it is fine to push then directly on the PR

I like it, thanks :).
You can always directly push code changes into my PRs, i have no problems with that.

> Why are the rectangles filled with black?

This is because of used color as `MapSetType.VIRTUAL_WALLS: "#f00"`
I have update it and also adjust alpha channel for a better look.

image

MVladislav:
By side i checked how `v2` is performing and found some "problems"/points I would need some help/decisions:

- The new version is named as `getMapSet_V2`, but i found a line which overwrites `_V2` inside `deebot_client/messages/__init__.py` in line `34-35`
  - This was a hell to understand why the events always use `getMapSet` instead of `getMapSet_V2`
  - ```py
    # T8 series and newer
    if converted_name.endswith("_V2"):
        converted_name = converted_name[:-3]
    ```
  - How should we proceed with that?
    - A possible solution could be, instead of create `getMapSet_V2`, we can combine it inside `getMapSet` with checks what data ins included, but not sure if this will work.
    - Or make a whitelist, for which command the `V2` should not be removed.
- I have update `getMapSet_V2` a bit:
  - When it is called to use `getMapSet_V2` directly, over `GetCachedMapInfo`, it is working as expected.
  - But when it is called by `onMapSet_V2`-"event/atr" it has less info and the map will not be updated
    - This is intercepted by `json/map.py` in line `301` by check if `subsets` is given
    - I would somehow call `GetMapSetV2` again to collect the needed information, like it is also done in `GetMapTrace`
      - where it calls itself inside `_handle_response`
      - but i'm not sure how this logic works
      - current when `GetMapSetV2` is called by `GetCachedMapInfo` it works and the map is updated
      - but when it is called by event/atr it will not be updated
  - We could also complete ignore that topic, to handle update the map by event/atr calls, as this also not works for older models where it should be updated by `GetMapSubSet`
    - changed virtual walls will than only load on inital calls not on events

MVladislav:
What do you think about, by extending the map with names?
(I **not pushed** this change now)

As Example:
image

MVladislav:
For the new bots in V2 API the walls are separate drawn, for that the event  `onMapInfo_V2` needs to be added.

In the app, the walls can be activated and deactivated, which will looks like:

|activated walls| deactivated walls|
|:-|:-|
|image|image|

Not sure if this is really needed, i have local a version which is half ready, but i get some errors which i can currently not fix. If this is something required, i can try to finish it, else i would remove it.

It would be looks as follows:

|without wall|with walls|
|:-|:-|
|image|image|

edenhaus:
> What do you think about, by extending the map with names, or maybe also with ID info?

Please don't add it in this PR. We should discuss this first via Discord

edenhaus:
>     * When it is called to use `getMapSet_V2` directly, over `GetCachedMapInfo`, it is working as expected.

This is how it should be

>     * But when it is called by `onMapSet_V2`-"event/atr" it has less info and the map will not be updated

Please implement `onMapSet_V2` like the `onBattery` in https://github.com/DeebotUniverse/client.py/blob/dev/deebot_client/messages/json/battery.py
See also https://github.com/DeebotUniverse/client.py/blob/dev/deebot_client/commands/json/battery.py to avoid duplicated code

By implementing the message `if converted_name.endswith("_V2"):` is no problem anymore and the way to go... I want to remove soon or later all the conventions fallbacks

edenhaus:
> n the app, the walls can be activated and deactivated, which will looks like:

We should keep this PR as small as possible and not add any more features... These features can be added in follow up PRs

MVladislav:
>>  What do you think about, by extending the map with names, or maybe also with ID info?
>
> Please don't add it in this PR. We should discuss this first via Discord
>
>> n the app, the walls can be activated and deactivated, which will looks like:
>
> We should keep this PR as small as possible and not add any more features... These features can be added in follow up PRs

I will let that changes be backup tracked in [issue for titles and walls](https://github.com/MVladislav/client.py/commit/cd47c6f4d1fbf712186ec4e1beda9d815bd813e9), if the code will be needed.

deebot_client/commands/json/map.py
```diff
@@ -246,6 +267,27 @@ def _handle_body_data_dict(
...
+    name = "getMapSet_V2"
+
+    def __init__(
+        self,
+        mid: str,
+        type: (  # pylint: disable=redefined-builtin
+            MapSetType | str
+        ) = MapSetType.ROOMS,
+        *,
+        is_available_check: bool = False,
```
edenhaus, line 282:
```suggestion
```

deebot_client/commands/json/map.py
```diff
@@ -246,6 +267,27 @@ def _handle_body_data_dict(
...
+    ) -> None:
+        if isinstance(type, MapSetType):
+            type = type.value
+
+        super().__init__({"mid": mid, "type": type})
+        self._is_available_check = is_available_check
```
edenhaus, line 288:
```suggestion
```

deebot_client/map.py
```diff
@@ -352,7 +334,7 @@ def _get_svg_subset(
 
     # For any other points count, return a polygon that should fit any required shape
     return svg.Polygon(
-        fill=_COLORS[subset.type] + "90",  # Set alpha channel to 90 for fill color
+        fill=_COLORS[subset.type] + "30",  # Set alpha channel to 90 for fill color
```
edenhaus, line 337:
```suggestion
        fill=_COLORS[subset.type] + "30",  # Set alpha channel to 30 for fill color
```

deebot_client/messages/json/map.py
```diff
@@ -0,0 +1,96 @@
...
+        if not data.get("subsets") and data.get("mid"):
+            # NOTE: here would be needed to call 'GetMapSetV2' again with 'mid' and 'type',
+            #       that on event will update the map set changes,
+            #       messages current cannot call commands again
+            return HandlingResult(
+                HandlingState.SUCCESS, {"mid": data["mid"], "type": data["type"]}
+            )
+
+        # subset is based64 7z compressed
+        subsets: list[list[str]] = json.loads(
```
edenhaus, line 42: If subsets are returned only by the get command than this code should be part of the command class and not message

tests/commands/json/test_map.py
```diff
@@ -22,39 +25,6 @@
 from . import assert_command
 
 
-async def test_getMapSubSet_customName() -> None:
```
edenhaus, line 25: This test tests the custom map name, and we should keep it.

MVladislav: true, to much revert/removed

deebot_client/map.py
MVladislav: updated

deebot_client/commands/json/map.py
MVladislav: removed

deebot_client/commands/json/map.py
MVladislav: removed

deebot_client/commands/json/map.py
```diff
@@ -246,6 +269,93 @@ def _handle_body_data_dict(
...
+            room_subsets: list[dict[str, Any]] = [
+                {
+                    "id": int(subset[0]),  # room id
+                    "name": subset[1]
+                    if subset[1] and subset[1] != " "
+                    else "Default",  # room name
+                    # subset[2] not sure what the value is for
+                    # subset[3] not sure what the value is for
+                    # subset[4] room clean order
+                    "coordinates": f"{subset[5]},{subset[6]}",  # room center coordinates
```
edenhaus, line 317: The coordinates on `GetMapSet` are the points/border of the specific room. These coordinates can be used to determine if the bot is in a specific room (https://github.com/DeebotUniverse/client.py/pull/397).

Does the V2 also have these values? At least, it looks like if you check out the PR.
We can use the center coordinates later in a follow up PR

edenhaus (changes requested):

deebot_client/commands/json/map.py
MVladislav: yes, the coordinate is the center of a room for newer bots.
For older bots i saw they use points of the room boarder.

You could not check for new bots this way if they are in a specific room as we only have the center of the room now.
For newer bots we get info for different room over the image, as they have now possibility for different coloring. It can be colored different by add colors in numbers higher than 100 inside:
- https://github.com/MVladislav/client.py/blob/9d203bab078b1fa499fd8798e40d6803a226445a/deebot_client/map.py#L127
- https://github.com/MVladislav/client.py/blob/9d203bab078b1fa499fd8798e40d6803a226445a/deebot_client/map.py#L346

I have had a test for set a room title in the map center, what was tested here: https://github.com/MVladislav/client.py/blob/cd47c6f4d1fbf712186ec4e1beda9d815bd813e9/deebot_client/map.py#L390

edenhaus (approved):

tobiasbrage:
Is a fix in the works? I'm really regretting buying an Ecovacs and not a Roborock at this point.

PR #376, opened 2023-12-18
Clean pads interval (wash interval)

NEVdataDyne:
Adds the option to set the interval between two washing of the clean pads.

I still encounter an issue that I can't solve and need someone who knows how to correct it :

Logger: custom_components.deebot.controller
Source: custom_components/deebot/controller.py:78
Integration: Deebot 4 Home Assistant ([documentation](https://deebot.readthedocs.io/integrations/home-assistant), [issues](https://github.com/DeebotUniverse/Deebot-4-Home-Assistant/issues))
First occurred: 17:30:55 (1 occurrences)
Last logged: 17:30:55

Error during setup
Traceback (most recent call last):
  File "/config/custom_components/deebot/controller.py", line 78, in initialize
    devices = await self._api_client.get_devices()
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/deebot_client/api_client.py", line 35, in get_devices
    static_device_info = get_static_device_info(device["class"])
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/deebot_client/hardware/deebot/__init__.py", line 27, in get_static_device_info
    _load()
  File "/usr/local/lib/python3.11/site-packages/deebot_client/hardware/deebot/__init__.py", line 21, in _load
    importlib.import_module(full_package_name)
  File "/usr/local/lib/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1147, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 690, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 940, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/usr/local/lib/python3.11/site-packages/deebot_client/hardware/deebot/85nbtp.py", line 121, in <module>
    get=[GetWashInterval()],
         ^^^^^^^^^^^^^^^^^
[...]

PR #387, opened 2024-01-05
Add support for set/getWashInfo commands and onWashInfo message

lukakama:
On T20 Omni it is possible to control mops washing mode (with cold or hot water) and level (standard or deep washing).

These features are handled by getWashInfo and setWashInfo commands, with a push onWashInfo message which is sent when settings are changed.

The `getWashInfo` returns 3 different information:
- `mode[int]`: this is the wash mode; it corresponds to "hot water cleaning" setting in the app, which can be disabled (value `0`) or enabled (value `1`)
- `hot_wash_amount[int]`: this is the washing level of mops, as cycles of 3 water charges and discharges on the station; the app allows to select only 2 values, one for "standard" washing (value `1` - 3 water charges and discharges) and one for deep washing (value `3` - 9 water charges and discharges).
- `interval[int]`: this is the interval between two mop washings while mopping.

The app uses the `setWashInfo` command to distrincly set mops washing mode and level, passing only `mode` or `hot_wash_amount` with the desired value. The interval, instead, is set using a differente command `setWashInterval` (handled in the PR #376).

Performing a `setWashInfo` command results in a `onWashInfo` message to be received from the MQTT server (with all the information of the `getWashInfo` command), regardless if `setWashInfo` is invoked passing only `mode` or only `hot_wash_amount`. The `onWashInfo` message is sent also when is performd a `setWashInterval` command.


lukakama (commented):
Currently I created 2 distinct commands (and thus two capabilities), one to set the only the `mode` and another to set only the `hot_wash_amount`, in order to mimic the app which handles them distinctly, but they could be merged into a single command that, however, should then send only one parameter between `mode` and `hot_wash_amount` when executed.

Also, due this asymmetry between get command (which returns 3 attributes) and set one (which support only 1, or 2 in case of merging), I had difficulties implementing set command tests. Any help is welcome :) .

lukakama:
Additionally, I used an Enum for `mode`, as, IMHO, the name seems much more related to a "washing mode" than to a "hot water washing on/off" option, where currently it is supported only "standard washing" and "hot water washing", and so, for future proof compatibility, an Enum allows to easily handle additional values. 
However, for the sake of simplicity, it can be changed to a `bool` "hot water on/off" value, but then, in case of future additional modes, a library (and library usages) refactor will be needed.

lukakama:
UPDATE: I did some additional tests and I can tell that, when `mode` is set to `0`, mops are **always** washed using a single wash cycle (3 charges and discharges of the station), regardless the value of `hot_wash_amount`, so it resemble even more that `mode` set to `0` correspond to a "Standard washing" (cold water with 1 cycle), and set to `1` correspond to a "Hot water washing" (Hot water with a configurable number of cycles).

lukakama:
UPDATE: I just found that the library does not support multiple command classes mapped to the same MQTT command in p2p commands handling: 
https://github.com/DeebotUniverse/client.py/blob/bb19d12dcbff706985ac651787670b488d219435/deebot_client/mqtt_client.py#L259-L270

As it can be seen, the code look for a command class mapped to the MQTT command topic, but in the PR there are 2 classes, `SetWashInfoMode` and `SetWashInfoHotWashAmount`,  mapped to the same `setWashInfo` MQTT command topic, and only one of them is added to the `COMMANDS_WITH_MQTT_P2P_HANDLING` mapping dictionary (I think that it depends on the order of class loading, where the last command loaded overwrite the previous and become the one used to handle p2p commands for that type). This then lead the code to raise an error when handling p2p `setWashInfo` commands when parameters does not match ones of the command class mapped into `COMMANDS_WITH_MQTT_P2P_HANDLING` in this method, as it would have only one of `mode` and `hot_wash_amount` parameters:
https://github.com/DeebotUniverse/client.py/blob/bb19d12dcbff706985ac651787670b488d219435/deebot_client/command.py#L274-L284

@edenhaus I don't know how to proceed... It seems that to support this setWashInfo command with different parameters it is required some sort of refactor on other parts of the library  which are currently quite complex for me to identify, and that will also depend on how this command handling will be implemented, between using two command classes vs using a single command class that, however, will have two mutually exclusive parameters... Maybe we should talk about it on discord to identify how this can be managed.

deebot_client/commands/json/wash_info.py
```diff
@@ -0,0 +1,66 @@
...
+    name = "setWashInfo"
+    get_command = GetWashInfo
+    _mqtt_params = MappingProxyType(
+        {
+            "hot_wash_amount": InitParam(int),
+        }
+    )
+
+    def __init__(self, hot_wash_amount: int) -> None:
+        super().__init__({"hot_wash_amount": hot_wash_amount})
```
edenhaus, line 66: This should be only one Command

```suggestion
class SetWashInfo(JsonSetCommand):
    """Set wash info command."""

    name = "setWashInfo"
    get_command = GetWashInfo
    _mqtt_params = MappingProxyType(
        {
            "mode": InitParam(int),
            "hot_wash_amount": InitParam(int),
        }
    )

    def __init__(self, mode: WashMode | str| None = None, hot_wash_amount: int | None) -> None:
        args = {}
        if isinstance(mode, str):
            mode = WashMode.get(mode)
        
        if mode is not None:
            args["mode"] = mode
            
        if hot_wash_amount is not None:
            agrs["hot_wash_amount"] = hot_wash_amount
        super().__init__(args)
```

deebot_client/events/wash_info.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+@dataclass(frozen=True)
+class WashInfoEvent(Event):
+    """Wash info event representation."""
+
+    mode: WashMode
+    # None means no data available
+    interval: int
+    hot_wash_amount: int
```
edenhaus, line 23:
```suggestion
    interval: int | None
    hot_wash_amount: int | None
```

If None is a valid value please also add it to typing

deebot_client/commands/json/wash_info.py
```diff
@@ -0,0 +1,66 @@
...
+
+from deebot_client.command import InitParam
+from deebot_client.event_bus import EventBus
+from deebot_client.events import WashInfoEvent, WashMode
+from deebot_client.message import HandlingResult
+
+from .common import JsonGetCommand, JsonSetCommand
+
+
+class GetWashInfo(JsonGetCommand):
```
edenhaus, line 14:
```suggestion
class GetWashInfo(OnWaterInfo, JsonGetCommand):
```

We don't need to duplicate the code

deebot_client/commands/json/wash_info.py
```diff
@@ -0,0 +1,66 @@
...
+        :return: A message response
+        """
+        event_bus.notify(
+            WashInfoEvent(
+                mode=WashMode(int(data["mode"])),
+                hot_wash_amount=data["hot_wash_amount"],
+                interval=data["interval"],
+            )
+        )
+        return HandlingResult.success()
```
edenhaus, line 34:
```suggestion
```

edenhaus (changes requested):
Sorry for the delay. It was a long month by porting the custom integration to core

edenhaus:
I think by merging both commands together I solved some issues you had.
When executing the command the caller can decide to pass both or only one argument

deebot_client/events/wash_info.py
lukakama: Sorry, I forgot to remove the comment: it comes from WaterInfoEvent class, which I used as template to code WashInfoEvent. Anyway it is not a bad idea to support also None types (I'm not confident enough that interval and hot_wash_amount are always present)

lukakama:
@edenhaus I refactored the code to work with a single command class, and thus I added a couple of tests for command execution, but it seems that the standard test `assert_set_command` does not handle well command with possible missing mqtt parameters (it rely on 1:1 relation between comand args and mqtt parameters). I don't know if it is just a test issue, or if it is something that could create issue at runtime. Do you have any hit on it? In case I could just change the test code in order to skip mqtt payload test.

edenhaus:
I fixed it for you.
First I added a default value to InitParam and afterwards I fixed the new command and the test for it

PR #397, opened 2024-01-18
Added room name information to Position object

francescopalma86:


deebot_client/command.py
```diff
@@ -94,11 +94,14 @@ async def execute(
...
 
     async def _execute(
-        self, authenticator: Authenticator, device_info: DeviceInfo, event_bus: EventBus
+        self, authenticator: Authenticator, device_info: DeviceInfo, event_bus: EventBus, rooms: list[Room] = []
     ) -> CommandResult:
         """Execute command."""
         response = await self._execute_api_request(authenticator, device_info)
 
+        if self.name == "getPos":
+            response["resp"]["body"]["data"]["rooms"] = rooms
```
edenhaus, line 103: Why is this needed?
`getPos` specific code should go into the `GetPos` class and not in the general `Command` class.

deebot_client/device.py
```diff
@@ -80,6 +85,12 @@ async def on_pos(event: PositionsEvent) -> None:
 
         self.events.subscribe(PositionsEvent, on_pos)
 
+        async def on_rooms(event: RoomsEvent) -> None:
```
edenhaus, line 88: I think this should go into `map.py`

deebot_client/map.py
```diff
@@ -407,7 +408,10 @@ async def on_map_set(event: MapSetEvent) -> None:
 
         async def on_map_subset(event: MapSubsetEvent) -> None:
             if event.type == MapSetType.ROOMS and event.name:
-                room = Room(event.name, event.id, event.coordinates)
+                decompressed_coords = _decompress_7z_base64_data(event.coordinates).decode('utf-8')
```
edenhaus, line 411: Compressed coordinates are Part of `GetMapSubset_V2` but not the `GetMapSubset` and therefore should be in the command class otherwise this code will break older bots

requirements.txt
```diff
@@ -5,3 +5,4 @@ defusedxml
 numpy>=1.23.2,<2.0
 Pillow>=10.0.1,<11.0
 svg.py>=1.4.2
+shapely~=2.0.2
```
edenhaus, line 8: Please loosen the requirements similar to the other requirements

edenhaus (changes requested):
Please make sure the CI is passing

deebot_client/command.py
francescopalma86: I don't know how to pass rooms parameter to GetPos class. This is the only way I've found. Will you give me a hand?

requirements.txt
francescopalma86: ok

deebot_client/command.py
edenhaus: I think this code should be in `map.py` and not in the command. The commands module is what we get from the API and we should not do calculations there.
Instead I think the calculation should be part of `map.py`, where you already have the rooms and the position

PR #405, opened 2024-01-28
Add support for Deebot OZMO T8+

markus99:
Appears this unit has the same features as the Ozmo T8 AIVI.  The main difference is the AIVI has a camera that can be accessed via the Deebot APP.  Doesn't appear this camera is referenced / used in the integration, so all other features are the same.  Created a link to the existing config vs. duplicating.  

Unsure if camera functionality (in HA especially) is a future consideration / feature request.  If so, I can update this new file w/ the entire current contents of deebot_client/hardware/deebot/x5d34r.py vs. just referencing.


deebot_client/hardware/deebot/55aiho.py
```diff
@@ -0,0 +1 @@
+x5d34r.py
```
edenhaus, line 1: Please create a symbol link to the `x5d34r.py` file. It looks like this is a text file with the name in it

edenhaus (changes requested):

PR #426, opened 2024-02-18
refactor - on-events - test cases

MVladislav:
- move some static values into const.py
- update code syntax
- update imports
- update/extend get-command by combine with on-messages
- fix comments
- update/extend test cases
- project config files updated


deebot_client/messages/json/map.py
```diff
@@ -33,3 +58,50 @@ def _handle_body_data_dict(
...
+
+    @classmethod
+    def _handle_body_data_dict(
+        cls, event_bus: EventBus, data: dict[str, Any]
+    ) -> HandlingResult:
+        """Handle message->body->data and notify the correct event subscribers.
+
+        :return: A message response
+        """
+        # onMinorMap sends no type, so fallback to "ol"
```
edenhaus, line 102: From this comment, it looks like `getMinorMap` and `onMinorMap` are sending different data. This function should only have the code for the message. Code for `getMinorMap` should be part of the `GetMinorMap` class

deebot_client/messages/json/pos.py
```diff
@@ -0,0 +1,46 @@
...
+    def _handle_body_data_dict(
+        cls, event_bus: EventBus, data: dict[str, Any]
+    ) -> HandlingResult:
+        """Handle message->body->data and notify the correct event subscribers.
+
+        :return: A message response
+        """
+        positions = []
+
+        for type_str in ["deebotPos", "chargePos"]:
```
edenhaus, line 28: Are we sure that we are getting multiple positions on a `onPos` message? I think it's only possible on the `getPos` command and `onPos` is only providing a single position.

edenhaus (changes requested):
Please split this PR into multiple PRs as it is too big to review. Aslo there are some merge conflicts

PR #432, opened 2024-02-20
Add Goat G1

Augar:
Adds mower capability


deebot_client/capabilities.py
```diff
@@ -121,6 +121,14 @@ class CapabilityCleanAction:
     area: Callable[[CleanMode, str, int], Command]
 
 
+@dataclass(frozen=True, kw_only=True)
```
edenhaus, line 124: Can we use generics to avoid duplicated code?

deebot_client/capabilities.py
```diff
@@ -133,6 +141,16 @@ class CapabilityClean:
     work_mode: CapabilitySetTypes[WorkModeEvent, WorkMode] | None = None
 
 
+@dataclass(frozen=True, kw_only=True)
+class CapabilityMow:
+    """Capabilities for mow."""
+
+    action: CapabilityMowAction
```
edenhaus, line 148: Here too

edenhaus (commented):
Thanks for the contributions :)

edenhaus (approved):
Thanks @Augar :+1:

PR #455, opened 2024-03-21
Readme: adjust example to make it work

flubshi:
This makes some changes to make the code example in Readme work.


edenhaus (approved):
Thanks

PR #457, opened 2024-03-24
Add xml command "GetFanSpeed"

flubshi:


flubshi:
pre-commit.ci autofix

deebot_client/events/fan_speed.py
```diff
@@ -17,6 +17,8 @@ class FanSpeedLevel(DisplayNameIntEnum):
     NORMAL = 0
     MAX = 1
     MAX_PLUS = 2
+    STANDARD = 3
+    STRONG = 4
```
edenhaus, line 21:
```suggestion
```

The JSON-based vacuums use the numbers to map to the correct value. Therefore we cannot introduce new ones in this enum.

Also, let's reuse the values already defined here. I would suggest we map standard to normal and strong to max. What do you think?

flubshi: I agree. It makes sense to map it to existing values. I updated this PR accordingly

flubshi:
pre-commit.ci autofix

flubshi:
pre-commit.ci autofix

edenhaus (approved):
Thanks @flubshi :+1:

PR #465, opened 2024-04-11
New goat g1 variant itk04l

Augar:
Add new goat g1 variant itk04l assuming it has the same capabilities as the initial goat g1 variant 5xu9h3.

Differences should only be of hull color and batteries size.


edenhaus (approved):
Thanks @Augar

PR #466, opened 2024-04-11
Fix nested f-strings

Augar:


Augar:
Only relevant for python 3.11
this repo uses 3.12 therefore not relevant

PR #467, opened 2024-04-11
Use IntEnum instead of DisplayNameIntEnum for CleanJobStatus

Augar:
We can use translations in HA and therefore, we don't need DisplayNameIntEnum anymore.
Remove it from CleanJobStatus


deebot_client/events/__init__.py
```diff
@@ -67,15 +66,15 @@ class BatteryEvent(Event):
...
+class CleanJobStatus(IntEnum):
     """Enum of the different clean job status."""
 
     NO_STATUS = -2
     CLEANING = -1
     # below the identified stop_reason values
     FINISHED = 1
-    MANUAL_STOPPED = 2, "manual stopped"
-    FINISHED_WITH_WARNINGS = 3, "finished with warnings"
+    MANUAL_STOPPED = 2
```
edenhaus, line 76:
```suggestion
    MANUALLY_STOPPED = 2
```

edenhaus (approved):
Thanks @Augar :+1:

PR #477, opened 2024-05-02
Update Deebot X2 Omni to use version 2 map commands

Ulfmerbold2000:
Since my X2 still wont show the map and rooms, i'll set it to Version 2 as a try


edenhaus:
@Ulfmerbold2000 Did you successfully test it? Is the map now working?

edenhaus:
pre-commit.ci autofix

PR #481, opened 2024-05-07
Added support for GOAT G1 800 (Grey)

Dolneaz:


edenhaus (changes requested):
Please make sure you use a symbol link to reference the other file. Currently it looks like it is a text file. That should also fix the CI

Dolneaz:
Should be fixed now. Man i hate this symlink stuff with Windows...

edenhaus (approved):

PR #490, opened 2024-05-10
Update error codes

Ulfmerbold2000:
Data actualized from
https://github.com/mrbungle64/ecovacs-deebot.js/blob/master/library/errorCodes.json

!!!Last Entry has SAME "," at the End as it was before the changes!!!

I have just added the newer errorcodes including that comma at the last entry. Since i've not familiar with python i choosed this way of editing instead deleting that last comma.


edenhaus:
pre-commit.ci autofix

PR #548, opened 2024-07-30
Add Deebot T30 Omni

pepperonzo:
Added work_mode capability for Deebot T30 Omni (by cloning T20 Omni capabilities)

## Summary by CodeRabbit

- **New Features**
	- Introduced a new module that enhances the Deebot hardware control system, potentially improving performance and expanding functionality.
- **Tests**
	- Added a test case for the new model identifier "z4lvk7," enhancing the robustness of the testing process.


pepperonzo:
Made some mess with commits due to recent updated libraries on main repo, but it should be fine now. Added the class `z4lvk7` to test_init.py as well

edenhaus:
> Made some mess with commits due to recently updated libraries on main repo, but it should be fine now. 
Will squash the PR so it's fine :)

PR #622, opened 2024-12-05
adding Deebot N30 Omni support (zwkcqc)

r1cht3r:
This is a relatively new model, this provides more features than I was previously getting with the fallback vacuum, including functioning work mode & clean count. Water flow rate was the only one that didn't work. 

## Summary by CodeRabbit

- **New Features**
  - Introduced a new robotic vacuum device configuration with advanced capabilities, including battery management, cleaning operations, fan speed adjustments, mapping, and sound playback.
  
- **Bug Fixes**
  - Expanded device testing coverage by adding the new device identifier and updating expected commands for existing device classes, enhancing the robustness of the testing framework.


edenhaus:
> This is a relatively new model, this provides more features than I was previously getting with the fallback vacuum, including functioning work mode & clean count. Water flow rate was the only one that didn't work.

That's the actual reason, why I removed the fallback so models get add with all supported features instead of only some

edenhaus (approved):
Thanks @r1cht3r :+1:

PR #644, opened 2024-12-07
Add Support for Deebot N8+ Black Variant

brendanm720:
Good Evening!

Apparently the white and black vaccums each have a different class ID.  I only have the White one at my disposal, and someone commented on my original pull request that theirs wasn't working with the log.

Please merge this file to add support for the Deebot N8+ Black Variant.  It is the same as the other one I submitted as they are identical vacuums other than the color of the plastic.

Thank you.

### DIFF
--- a/deebot_client/hardware/deebot/b2jqs4.py
+++ b/deebot_client/hardware/deebot/b2jqs4.py
@@ -1,0 +1,198 @@
+"""DEEBOT N8+ Black Capabilities."""
+
+from __future__ import annotations
+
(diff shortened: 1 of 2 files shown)


brendanm720:
Looks like the code checks failed because something wasn't imported when it ran.

flangio:
Hi,
can you also support N8 (not plus) black variant, model 7zya6u?
Thanks!

edenhaus (approved):
Thanks @brendanm720 :+1:

PR #645, opened 2024-12-08
Add base station functions

MarkGodwin:
The N20 Plus has a cyclone vacuum base station for emptying the deebot.

This comes with some an additional resettable counter for an internal filter inside the base station, which I added as a new resettable `Lifespan` type.

I also found a way to invoke an emptying of the Deebot, as well as an event that gets sent out during the automated vacuum process, which would maybe be useful to have if creating an automation to warn when the bin has been sucked out 20 times or something.

I couldn't see any existing Deebots that had any base-station functionality implemented, so I tried adding some of the configuration to allow this.

I'm a bit unsure as to what the implications of adding to the Capability classes really is - I'm sort of coding a bit blind, but it worked in a test script. Happy to change it further if this wasn't right.

## Release Notes

- **New Features**
  - Introduced base station functionalities, including actions and events.
  - Added support for handling station state messages.
  - Enhanced device capabilities for the Deebot N20 Plus, including battery management and cleaning actions.
  - Expanded command handling to include `StationAction` for base station operations.

- **Bug Fixes**
  - Enhanced error handling for device operations.

- **Tests**
  - Expanded test coverage for new message types related to base station actions.
  - Added tests for the new device identifier and its capabilities.

These updates enhance the device's operational capabilities, allowing for more efficient management and interaction with base station features.


MarkGodwin:
N20 Pro Plus was added recently - this should probably point here too, instead of to the N20.

MarkGodwin:
Ah - Just found #603 . Some of my changes need to be updated to fit into the new `station` capability set that was added there. I can do that once #603 is merged.
I think the new auto-empty settings added there are probably valid for the N20 Plus too.

edenhaus (approved):
Thanks @MarkGodwin :+1: 

I refactored the code a little bit and added tests for the new functionality

MarkGodwin:
I tried out your new changes with my N20 Plus, it all looks good. Thanks!
We've lost the setting for the Base station filter lifecycle parameter ("spHeap") in the config. Is that because the HA integration needs to be updated first?

PR #651, opened 2024-12-08
Add support for Deebot T9 (ipohi5)

WARIO2412:
Add Deebot T9 support, class 'ipohi5'.

My Deebot T9 has a different class than other ones. Therefore I added my class in this PR.
The simlink is also pointing to the Deebot T8 (x4d34r.py). Everything works as expected right now.


edenhaus (approved):
Thanks @WARIO2412 :+1:

Issue #555, opened 2024-08-08
T30 Omni new features (full support)


pepperonzo:
### Checks

- [X] I have searched the existing issues and no issue is describing my issue
- [X] I have checked the [FAQ](https://deebot.readthedocs.io/home/faq/)
- [X] I have checked the [documentation](https://deebot.readthedocs.io/integrations/home-assistant)
- [X] I have installed the latest version

### The problem

Home Assistant doesn't support all new features of Deebot T30 Omni (many of them in common with T20)
I've mapped them all, but I think I lack the knowledge to implement them by myself.
All I can do is a copy->paste->edit work and, for what I've seen, it's not enough.
Hope this can help

```
Mopping Plates Auto-Lifting	setCarpetInfo
auto-lifting		body:{"data":{"mode":0}}
bypass carpets		body:{"data":{"mode":1}}
no-lifting		body:{"data":{"mode":2}}
	
	
Do Not Disturb	setBlock
on		body:{"data":{"enable":1,"start":"22:0","end":"8:0"}}
off		body:{"data":{"enable":0,"start":"22:0","end":"8:0"}}
	
	
Child Lock	setChildLock
on		body:{"data":{"on":1}}
off		body:{"data":{"on":0}}
	
	
Foot Touch Control	setSwitchState
on		body:{"data":{"kickClean":1}}
off		body:{"data":{"kickClean":0}}
	
	
TrueDetect 3D Sensitivity	setTrueDetect
off		body:{"data":{"enable":0}}
on - standard	body:{"data":{"enable":1,"level":1}}
on- high	body:{"data":{"enable":1,"level":0}}
	
	
Mop Auto-Wash Frequency	setWashInfo
by time		body:{"data":{"roomwise":0}
by room		body:{"data":{"roomwise":1}}
Mop Auto-Wash Frequency - By time - Interval	setWashInterval
10 min		body:{"data":{"interval":10}}
15 min		body:{"data":{"interval":15}}
25 min		body:{"data":{"interval":25}}
	
	
Hot Air Drying Duration	setDryingDuration
2 hours		body:{"data":{"duration":120}}
3 hours		body:{"data":{"duration":180}}
4 hours		body:{"data":{"duration":240}}
	
	
Auto-Empty	setAutoEmpty
auto		body:{"data":{"enable":1,"frequency":"auto"}}
smart		body:{"data":{"enable":1,"frequency":"smart"}}
on		body:{"data":{"enable":1,"frequency":"auto"}
[...]

PR #659, opened 2024-12-10
Add Deebot X5 Pro Omni (lr4qcs)

XxInvictus:
- Initial clone of most compatible, requires extra features not supported yet in #555


n-andflash:
Hi @XxInvictus ,
My Deebot X5 Pro Omni has "rvflzn", not "lr4qcs". What is the difference? Should I PR this one?

스크린샷 2024-12-12 131713

XxInvictus:
@n-andflash yeah you should be able to use this as a template and symlink to the same file for your model id. Probably just a regional thing :)

PR #669, opened 2024-12-11
Add support for the N20 Plus

edenhaus:
@MarkGodwin added your model with this PR, so it works as for the new base station commands, I need some further testing


MarkGodwin:
Sure thing, @edenhaus. Happy to help with the testing if you need anything.

PR #677, opened 2024-12-13
Support mop autowash interval

XxInvictus:
Ref #555
- Add support for Mop Auto-Wash Frequency
- Add Capability to T20/T30 Omni/X5 Pro Omni via `p1jij8.py`

Tested/Validated with X5 Pro Omni:
image
image


XxInvictus:
Finally have all pre-commit.ci checks passing 🎉

edenhaus:
Can you confirm that the T20 (`p1jij8`) has this feature?

XxInvictus:
I can't confirm one way or the other, this isn't a feature which is advertised in marketing material so it's completely unknown. Given I can't say with 100% certainty happy to update to split T30 Omni and X5 Pro Omni into their own file to ensure there is no clash.

XxInvictus:
Looks like I just managed to confirm it, mentions it in [this review](https://www.androidpolice.com/ecovacs-deebot-t20-omni-review/)
> and lets you customize the vacuuming and mopping settings for each room, as well as the mop washing frequency

deebot_client/events/mop_auto_wash_frequency.py
```diff
@@ -0,0 +1,24 @@
...
+from enum import IntEnum, unique
+
+from .base import Event
+
+
+@unique
+class MopAutoWashFrequency(IntEnum):
+    """Enum class for all possible mop auto-wash frequencies."""
+
+    TEN_MINUTES = 10
```
edenhaus, line 15: As the values are numbers, can we also set the interval to 1 minute? And would the bot respect this set one minute?

XxInvictus: It doesn't seem like it, there may be some other magic values but these are the available presets for the X5 Pro Omni.

Padparadscha: In the App from T30 I can only chose between 10, 15 and 25 minutes for time based mop auto wash.
The other option is mop after room

deebot_client/device.py
```diff
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import asyncio
-from collections.abc import Callable, Coroutine
+from collections.abc import Callable, Coroutine  # pylint: disable=import-error
```
edenhaus:
```suggestion
from collections.abc import Callable, Coroutine
```

PR #679, opened 2024-12-13
Add X5 OMNI support (4jd37g)

chrfin:
Initial clone of most compatible based on #659


XxInvictus:
From my MR you referenced you should also add the model to the test suite at `tests/hardware/test_init.py` 😄

chrfin:
> From my MR you referenced you should also add the model to the test suite at `tests/hardware/test_init.py` 😄

Saw that, but in the current dev-branch this file looks different now and does not have that device list anymore...

edenhaus:
> From my MR you referenced you should also add the model to the test suite at `tests/hardware/test_init.py` 😄

I changed the test so it's not needed anymore as to many forgotten to update the tests :)

PR #806, opened 2025-02-05
Add support for Deebot T30S Care (63cum9)

worldserial:
Add support for Deebot T30S Care 63cum9 via symlink.


edenhaus (approved):
thanks @worldserial :+1:

PR #813, opened 2025-02-09
Add XML charge command

flubshi:
Implement new XML charge command: `<ctl><charge type='go'/></ctl>`

Runtime tested - works :)


edenhaus (approved):
Thanks @flubshi :+1:

PR #560, opened 2024-08-26
Add XML command "SetFanSpeed"

flubshi:
Adds support for deebot 900 XML command "SetFanSpeed". 

@edenhaus Could you please review?

Runtime tested:

image

## Summary by CodeRabbit

- **New Features**
	- Introduced enhanced command handling for XML messages.
	- Added commands for getting and setting clean speed, improving device operation control.
	- Expanded fan speed options with new levels: STANDARD and STRONG.

- **Bug Fixes**
	- Improved handling of command execution success and failure notifications.

- **Tests**
	- Added tests for the new clean speed commands to ensure expected behavior and outputs.

PR #817, opened 2025-02-10
Add XML command "CleanArea"

flubshi:
Our first XML clean command :tada: 

Tested by cleaning my kitchen and bathroom multiple times.

### DIFF
--- a/deebot_client/commands/xml/clean.py
+++ b/deebot_client/commands/xml/clean.py
@@ -1,0 +1,27 @@
+"""Clean commands."""
+
+from __future__ import annotations
+
+from deebot_client.models import CleanAction, CleanMode
+
+from .common import ExecuteCommand
+
+
+class CleanArea(ExecuteCommand):
+    """Clean area command."""
+
+    NAME = "Clean"
+    HAS_SUB_ELEMENT = True
+
+    def __init__(self, mode: CleanMode, area: str, cleanings: int = 1) -> None:
+        # <ctl><clean type='SpotArea' act='s' speed='standard' deep='1' mid='4,5'/></ctl>
+
+        super().__init__(
+            {
+                "type": mode.xml_value,
+                "act": CleanAction.START.xml_value,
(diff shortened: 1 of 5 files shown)


deebot_client/commands/xml/clean.py
```diff
@@ -0,0 +1,27 @@
...
+    HAS_SUB_ELEMENT = True
+
+    def __init__(self, mode: CleanMode, area: str, cleanings: int = 1) -> None:
+        # <ctl><clean type='SpotArea' act='s' speed='standard' deep='1' mid='4,5'/></ctl>
+
+        super().__init__(
+            {
+                "type": mode.xml_value,
+                "act": CleanAction.START.xml_value,
+                "speed": "standard",  # TODO: FanSpeedLevel.NORMAL.xml_value, after #560 is merged
```
edenhaus, line 23: Would this mean that the bot will always start with the fan level at `standard` and not the one defined in the fan speed entity?

deebot_client/util/enum.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def __new__(cls, value: str, xml_value: str = "") -> Self:
+        """Create new StrEnumWithXml."""
+        obj = str.__new__(cls, value)
+        obj._value_ = value
+        obj.xml_value = xml_value
+        return obj
+
+    @classmethod
+    def from_xml(cls, value: str) -> Self:
```
edenhaus, line 22: Can we add test for it?

edenhaus (commented):
I added a commit to reduce

deebot_client/util/enum.py
flubshi: Tests added in 7cca294fd3ad41e3f5c961e12ee9cf3e43d272e0

deebot_client/commands/xml/clean.py
flubshi: I think it makes sense to pass it as parameter. But we have to wait for #560 for the new FanSpeedLevel Enum.

We can either merge it like it is and address this in another PR, or we wait for #560

flubshi:
Closing in favor of #908

PR #827, opened 2025-02-19
Fix Deebot T10 (jtmf04) capabilities

PiperNigrum:
Fix of capability to match with t8 aivi


PiperNigrum:
@edenhaus: Hope that this is the right way to fix the wrong capability reference for the T10

PiperNigrum:
Are there any open task on my site? Because the pull request/merge process failed?

PiperNigrum (commented):
lookup for the file and found it - passed

PiperNigrum:
I'm not understanding why the run failed at mypy. The file is existing. I just changed the content.

image

edenhaus:
> I'm not understanding why the run failed at mypy. The file is existing. I just changed the content.

The file is a symbolic link and not a normal text file. Terefore it's not possible to change the file were it's pointing at online. Instead you need to delete it and recreate the symbolic link to the correct file.
I fixed it for you

PR #870, opened 2025-03-29
Add support for GOAT O500 Panorama (300lc5)

stephanerosi:
Simple copy of Goat G1 configuration
Tested with my Goat 0500 Panorama.
Works perfectly.


Zarox666:
Brandnew GOAT O500 Panorama here.
HA integration works to retrieve Information.
Controll wise clicking this works:
image
Clicking the Play, Pause or Return Button below does nothing.

When its docked and the blue Text says "Start Mowing" also nothing happens. 
So I can recall it home, but not send it out.

Support for the GOAT O500 would be very nice!
Its currently for sale for below 600€ and I guess thus will increase in popularity even further.

PR #893, opened 2025-04-19
Add legacy GetNetInfo JSON command

nanomad:
This is needed for Legacy XML bots such as Ozmo 905

### DIFF
--- a/deebot_client/commands/json/network.py
+++ b/deebot_client/commands/json/network.py
@@ -4,10 +4,15 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any
 
+from deebot_client.command import CommandWithMessageHandling
 from deebot_client.events import NetworkInfoEvent
-from deebot_client.message import HandlingResult, MessageBodyDataDict
+from deebot_client.message import (
+    HandlingResult,
+    MessageBodyDataDict,
+    MessageDict,
+)
 
-from .common import JsonCommandWithMessageHandling
+from .common import JsonCommand, JsonCommandWithMessageHandling
 
 if TYPE_CHECKING:
     from deebot_client.event_bus import EventBus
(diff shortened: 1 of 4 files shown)


deebot_client/commands/json/network.py
```diff
@@ -35,3 +38,27 @@ def _handle_body_data_dict(
             )
         )
         return HandlingResult.success()
+
+
+class GetNetInfoLegacy(JsonCommandWithRawMessageHandling):
```
edenhaus:
```suggestion
class GetNetInfoLegacy(JsonCommand, CommandWithMessageHandling, MessageDict):
```

As the legacy command is an exception and no other command uses the "raw" format, we don't need a class for it. We can always add it later if needed

deebot_client/commands/json/common.py
```diff
@@ -62,6 +63,12 @@ class JsonCommandWithMessageHandling(
     """Command, which handle response by itself."""
 
 
+class JsonCommandWithRawMessageHandling(
+    JsonCommand, CommandWithMessageHandling, MessageDict, ABC
+):
+    """Command, which handle raw response by itself."""
+
+
```
edenhaus:
```suggestion
```

deebot_client/message.py
```diff
@@ -140,7 +140,43 @@ def _handle(
...
+
+    @classmethod
+    def _handle(
+        cls, event_bus: EventBus, message: dict[str, Any] | str
+    ) -> HandlingResult:
+        """Handle message and notify the correct event subscribers.
+
+        :return: A message response
+        """
+        # This basically means an XML message
```
edenhaus:
```suggestion
```

This command is wrong

edenhaus (changes requested):

edenhaus (approved):
Thanks @nanomad :+1:

PR #896, opened 2025-04-19
Capabilities should not depend on JSON commands

nanomad:
This is mandatory as we are going to have ExecuteCommands in XML as well


edenhaus (approved):
Thanks @nanomad :+1:

PR #897, opened 2025-04-19
Checkout sh file as lf-terminated

nanomad:
Fixes windows devcontainer when the code is cloned from windows


edenhaus (approved):
Thanks @nanomad :+1:

PR #898, opened 2025-04-19
Add minor map to map capabilities

nanomad:
This is preparation for XML bots having map support

### DIFF
--- a/deebot_client/device.py
+++ b/deebot_client/device.py
@@ -38,7 +38,6 @@ if TYPE_CHECKING:
 _LOGGER = get_logger(__name__)
 _AVAILABLE_CHECK_INTERVAL = 60
 
-
 DeviceCommandExecute = Callable[[Command], Coroutine[Any, Any, dict[str, Any]]]
 
 
@@ -67,7 +66,13 @@ class Device:
             self.execute_command, self.capabilities.get_refresh_commands
         )
 
-        self.map: Final[Map] = Map(self.execute_command, self.events)
+        self.map: Final[Map] = Map(
+            self.execute_command,
+            self.events,
+            self.capabilities.map.minor.execute
+            if self.capabilities and self.capabilities.map
+            else None,
+        )
 
         async def on_pos(event: PositionsEvent) -> None:
             if self._state == StateEvent(State.DOCKED):
--- a/deebot_client/device.py
+++ b/deebot_client/device.py
@@ -67,11 +67,7 @@ class Device:
         )
 
         self.map: Final[Map] = Map(
-            self.execute_command,
-            self.events,
-            self.capabilities.map.minor.execute
-            if self.capabilities and self.capabilities.map
-            else None,
+            self.execute_command, self.events, self.capabilities.map
         )
 
         async def on_pos(event: PositionsEvent) -> None:
--- a/deebot_client/device.py
+++ b/deebot_client/device.py
@@ -66,8 +66,10 @@ class Device:
             self.execute_command, self.capabilities.get_refresh_commands
         )
 
-        self.map: Final[Map] = Map(
-            self.execute_command, self.events, self.capabilities.map
+        self.map: Final[Map | None] = (
+            Map(self.execute_command, self.events, self.capabilities.map)
+            if self.capabilities.map
+            else None
         )
 
         async def on_pos(event: PositionsEvent) -> None:
@@ -142,7 +144,8 @@ class Device:
                 await self._available_task
 
         await self.events.teardown()
-        await self.map.teardown()
+        if self.map:
+            await self.map.teardown()
 
     async def _available_task_worker(self) -> None:
         while True:
--- a/deebot_client/hardware/deebot/9s1s80.py
+++ b/deebot_client/hardware/deebot/9s1s80.py
@@ -42,11 +42,7 @@ from deebot_client.commands.json.efficiency import GetEfficiencyMode, SetEfficie
 from deebot_client.commands.json.error import GetError
 from deebot_client.commands.json.fan_speed import GetFanSpeed, SetFanSpeed
 from deebot_client.commands.json.life_span import GetLifeSpan, ResetLifeSpan
-from deebot_client.commands.json.map import (
-    GetCachedMapInfo,
-    GetMajorMap,
-    GetMapTrace,
-)
+import deebot_client.commands.json.map
 from deebot_client.commands.json.multimap_state import (
     GetMultimapState,
     SetMultimapState,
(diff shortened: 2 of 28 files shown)


deebot_client/hardware/deebot/9s1s80.py
```diff
@@ -164,17 +160,26 @@
         ),
         map=CapabilityMap(
             cached_info=CapabilityEvent(
-                CachedMapInfoEvent, [GetCachedMapInfo(version=2)]
+                CachedMapInfoEvent,
+                [deebot_client.commands.json.map.GetCachedMapInfo(version=2)],
```
edenhaus: Please revert all unrelated changes and import the Commands instead of the module

tests/hardware/test_init.py
```diff
@@ -27,7 +27,11 @@
 from deebot_client.commands.json.error import GetError
 from deebot_client.commands.json.fan_speed import GetFanSpeed
 from deebot_client.commands.json.life_span import GetLifeSpan
-from deebot_client.commands.json.map import GetCachedMapInfo, GetMajorMap, GetMapTrace
+from deebot_client.commands.json.map import (
```
edenhaus: Revert this changes

deebot_client/map.py
```diff
@@ -45,10 +44,14 @@ def __init__(
         self,
         execute_command: DeviceCommandExecute,
         event_bus: EventBus,
+        minor_map_command: Callable[[str, int], Command] | None,
```
edenhaus:
```suggestion
        capabilities: CapabilityMap,
```

deebot_client/map.py
```diff
@@ -45,10 +44,14 @@ def __init__(
...
     ) -> None:
         self._execute_command = execute_command
         self._event_bus = event_bus
 
+        self._minor_map_command: Callable[[str, int], Command] | None = (
+            minor_map_command
+        )
```
edenhaus:
```suggestion
        self._capabilities = capabilities
```

deebot_client/map.py
```diff
@@ -101,10 +104,11 @@ async def on_major_map(event: MajorMapEvent) -> None:
                     if (
                         self._map_data.map_piece_crc32_indicates_update(idx, value)
                         and event.requested
+                        and self._minor_map_command is not None
```
edenhaus:
```suggestion
```

deebot_client/map.py
```diff
@@ -101,10 +104,11 @@ async def on_major_map(event: MajorMapEvent) -> None:
...
                     ):
                         tg.create_task(
                             self._execute_command(
-                                GetMinorMap(map_id=event.map_id, piece_index=idx)
+                                self._minor_map_command(event.map_id, idx)
```
edenhaus:
```suggestion
                                self._capabilities.minor.execute(event.map_id, idx)
```

deebot_client/device.py
```diff
@@ -67,7 +66,13 @@ def __init__(
...
         )
 
-        self.map: Final[Map] = Map(self.execute_command, self.events)
+        self.map: Final[Map] = Map(
+            self.execute_command,
+            self.events,
+            self.capabilities.map.minor.execute
+            if self.capabilities and self.capabilities.map
+            else None,
+        )
```
edenhaus:
```suggestion
        self.map: Final[Map | None] = (
            Map(self.execute_command, self.events, self.capabilities.map)
            if self.capabilities.map
            else None
        )
```

edenhaus (changes requested):

deebot_client/map.py
edenhaus: Please implemented my requested changes

deebot_client/device.py
edenhaus: Same here

edenhaus (changes requested):
Please implement my requested changes. If you think there is a better approach than comment on the comment but don't resolve it without implementing the requested changes or any comment on it

nanomad:
I think I've missed a push after the rebase, I'll take a look at it
tomorrow

Il Mar 22 Apr 2025, 23:44 Robert Resch ***@***.***> ha
scritto:

nanomad:
> Please implement my requested changes. If you think there is a better approach than comment on the comment but don't resolve it without implementing the requested changes or any comment on it

Uhm, I'm looking at the code in the "files" section and I do see your changes integrated. I think github is potentially showing you outdated info?

deebot_client/map.py
nanomad: Already done?

deebot_client/device.py
nanomad: Done as well :)

edenhaus:
> Uhm, I'm looking at the code in the "files" section and I do see your changes integrated. I think github is potentially showing you outdated info?

grafik
I see this and here my requested changes are missing. I also checked out the code local and there also my requested changes are missing. Are you sure you have pushed all changes?

nanomad:
@edenhaus I do now see what you mean. I'll rework this PR as making map nullable requires other changes as well

nanomad:
@edenhaus Done, I guess this is a breaking change since
- We've added a new mandatory capability (low impact since this PR takes care of migrating existing bots as well)
- The map attribute of the device can now be None (instead of simply not working for bots such as 2ap5uq)

tests/conftest.py
```diff
@@ -157,6 +166,17 @@ def device_info(
...
+def device_info_no_map(
+    api_device_info: ApiDeviceInfo,
+    static_device_info_no_map: StaticDeviceInfo,
+) -> DeviceInfo:
+    return DeviceInfo(
+        api_device_info,
+        static_device_info_no_map,
+    )
+
+
```
edenhaus:
```suggestion
```

Not needed as we can overwrite the `static_device_info` fixture if needed

tests/conftest.py
```diff
@@ -129,6 +129,15 @@ async def test_mqtt_client(
...
     assert info is not None
+    assert info.capabilities.map is not None
+    return info
+
+
+@pytest.fixture
+async def static_device_info_no_map() -> StaticDeviceInfo:
+    info = await get_static_device_info("2ap5uq")
+    assert info is not None
+    assert info.capabilities.map is None
```
edenhaus:
```suggestion
```

Not needed here if used only by one test

tests/test_device.py
```diff
@@ -133,3 +133,15 @@ async def test_mac_address(
 
     assert device.mac == mac
     await device.teardown()
+
+
+async def test_behaviour_with_no_map_capability(
+    authenticator: Authenticator, device_info_no_map: DeviceInfo
+) -> None:
+    device = Device(device_info_no_map, authenticator)
```
edenhaus:
```suggestion
def static_device_info_no_map() -> StaticDeviceInfo:
    """Return a StaticDeviceInfo without map capability."""
    info = asyncio.run(get_static_device_info("2ap5uq"))
    assert info is not None
    assert info.capabilities.map is None
    return info


@pytest.mark.parametrize(
    "static_device_info",
    [
        static_device_info_no_map(),
    ],
)
async def test_behaviour_with_no_map_capability(
    authenticator: Authenticator, device_info: DeviceInfo
) -> None:
    device = Device(device_info, authenticator)
```

tests/test_device.py
```diff
@@ -133,3 +133,15 @@ async def test_mac_address(
...
+    # deactivate refresh event subscribe refresh calls
+    device.events._get_refresh_commands = lambda _: []
```
edenhaus: Why?

edenhaus (changes requested):

edenhaus (approved):
Thanks @nanomad :+1:

PR #900, opened 2025-04-19
Add XML Message BatteryInfo

nanomad:


deebot_client/messages/__init__.py
```diff
@@ -53,7 +55,7 @@ def get_message(message_name: str, data_type: DataType) -> type[Message] | None:
     If there exists no exact match, some conversations are performed on the name to get message object similar to the name.
     """
     messages = MESSAGES.get(data_type)
-    if messages is None:
+    if messages is None:  # pragma: no cover
```
edenhaus:
```suggestion
    if messages is None:
```

Never exclude code from the coverage report

tests/messages/__init__.py
```diff
@@ -11,11 +11,20 @@
...
 ) -> None:
     event_bus = Mock(spec_set=EventBus)
 
     result = message.handle(event_bus, data)
 
     assert result.state == HandlingState.SUCCESS
     event_bus.notify.assert_called_once_with(expected_event)
+
+
+def assert_message_failure(message: type[Message], data: dict[str, Any] | str) -> None:
```
edenhaus:
```suggestion
def assert_message_failure(message: type[Message], data: dict[str, Any] | str, expected_result_state: HandlingState) -> None:
```

tests/messages/__init__.py
```diff
@@ -11,11 +11,20 @@
...
+    event_bus = Mock(spec_set=EventBus)
+
+    result = message.handle(event_bus, data)
+
+    assert result.state != HandlingState.SUCCESS
```
edenhaus:
```suggestion
    assert result.state == expected_result_state
```

edenhaus (changes requested):

edenhaus (approved):
Thanks @nanomad :+1:

PR #902, opened 2025-04-19
Make volume capability optional.

nanomad:
This is needed as OZMO 900 and 905 have no volume setting 

**Note: This is a breaking change**


edenhaus:
This PR and also #903 can be done directly when adding the Ozmo 905 as they are really small and I also don't want to add a breaking change without any usage.

nanomad:
Drafting this PR just to avoid forgetting about it

PR #903, opened 2025-04-19
Make Multimap support optional.

nanomad:
This is needed as OZMO 905 has no Multimap support.

NOTE: This is a breaking change


nanomad:
Drafting this PR just to avoid forgetting about it

PR #909, opened 2025-04-19
Add XML Map commands

nanomad:


deebot_client/commands/xml/map.py
```diff
@@ -0,0 +1,281 @@
...
+
+    _ARGS_MSID = "msid"
+    _ARGS_TYPE = "type"
+    _ARGS_SUBSETS = "subsets"
+
+    NAME = "GetMapSet"
+
+    def __init__(
+        self,
+        # pylint: disable=unused-argument
```
edenhaus: Why did you add an unused param?

nanomad: It's a piece of old code, but J've just realized we are missing the code that triggers a GetMapSet from GetMapSt

nanomad: @edenhaus done, I've also made sure GetMapSt calls GetMapSet automatically as it does for the JSON variant

edenhaus (approved):
Thanks @nanomad :+1:

PR #910, opened 2025-04-20
Implement XML Command GetCleanLogs

nanomad:


deebot_client/commands/xml/enum.py
```diff
@@ -0,0 +1,42 @@
...
+from __future__ import annotations
+
+from enum import StrEnum, unique
+from typing import Self
+
+from deebot_client.events import CleanJobStatus
+
+
+@unique
+class XmlStopReason(StrEnum):
```
edenhaus: Please move this enum to clean_logs.py as it's used only there

tests/commands/xml/test_enum.py
```diff
@@ -0,0 +1,15 @@
+from __future__ import annotations
```
edenhaus: We can delete this test as we test it implicit with the above one

edenhaus (changes requested):

tests/commands/xml/test_enum.py
nanomad: Done

edenhaus (approved):
Thanks @nanomad :+1:

PR #915, opened 2025-04-22
Implement XML ChargeState messages

nanomad:


deebot_client/commands/xml/charge_state.py
```diff
@@ -27,14 +27,13 @@ def _handle_xml(cls, event_bus: EventBus, xml: Element) -> HandlingResult:
```
edenhaus: To avoid duplicate code, please remove this function here and add the CargeState to the parent classes:

```python
class GetChargeState(XmlCommandWithMessageHandling, ChargeState):
```

edenhaus (changes requested):

deebot_client/commands/xml/charge_state.py
nanomad: XML messages have no "ret" attribute though, how about an ABC that both GetChargeState and ChargeState inherit from so that the charge sub-element parsing stays the same?

edenhaus: Fine for me :) I just want to avoid duplicated code

nanomad: @edenhaus take a look at the new version

edenhaus (approved):

edenhaus:
Please fix the merge conflicts

nanomad:
@edenhaus Rebased but CI  needs restarting I think

edenhaus (approved):
Thanks @nanomad :+1:

PR #916, opened 2025-04-22
Implement XML Position messages

nanomad:


tests/messages/xml/test_pos.py
```diff
@@ -0,0 +1,30 @@
...
+from deebot_client.message import HandlingState
+from deebot_client.messages.xml import Pos
+from deebot_client.rs.map import PositionType
+from tests.messages import assert_message, assert_message_failure
+
+
+@pytest.mark.parametrize("position", [(-9, 15, 89)])
+def test_Pos(position: tuple[int, int, int]) -> None:
+    x, y, a = position
+    xml_message = f'<ctl td="Pos" t="p" p="{x},{y}" a="{a}" valid="1" />'
```
edenhaus, line 15: Some questions about the message:
- is `t` the type and will it be different for the charger position?
- There is a `valid` attribute... I would assume that the position is only valid if it is equal to 1. Can you please verify it

deebot_client/messages/xml/pos.py
```diff
@@ -0,0 +1,34 @@
...
+
+    NAME = "Pos"
+
+    @classmethod
+    def _handle_xml(cls, event_bus: EventBus, xml: Element) -> HandlingResult:
+        if p := xml.attrib.get("p"):
+            p_x, p_y = p.split(",", 2)
+            p_a = xml.attrib.get("a", 0)
+            position = Position(
+                type=PositionType.DEEBOT, x=int(p_x), y=int(p_y), a=int(p_a)
```
edenhaus: Is the app not showing where the charger is?

tests/messages/xml/test_pos.py
nanomad: Fair point
- I have no way of checking if valid = 0 ever happens, but I guess one more check doesn't hurt.
- The app does show the charger position, I just didn't figure out what param it uses. I'll check that tomorrow

nanomad: There's a GetChargerPos command, but I didn't find any "Pos" events for it.
Suggestion:
- Imlpement GetChargerPos 
- Only handle valid="1" t="p" messages as bot position, we'll leave everything else as HandlingResult.analyse() so that they can be eventually debugged further

deebot_client/messages/xml/pos.py
flubshi: For Deebot 900 it shows the charger position in app.

But it looks like the position is explicitly requested:

```
Got message: topic=iot/p2p/GetChargerPos/HelperMQClientId-awseu-sts-ngiot-mqsjmq-43/ecosys/<id>/<uuid>/ls1ok3/<..>/q/<...>/x, payload=b'<ctl/>'
[...]
Got message: topic=iot/p2p/GetChargerPos/<uuid>/ls1ok3/<id>/HelperMQClientId-awseu-sts-ngiot-mqsjmq-43/ecosys/<..>/p/<...>/x, payload=b"<ctl ret='ok' p='-200,816' a='-90'/>"
```

Edit: ahh, noticed that @nanomad already replied in the other comment

deebot_client/commands/xml/pos.py
```diff
@@ -16,7 +16,25 @@
     from deebot_client.event_bus import EventBus
 
 
-class GetPos(XmlCommandWithMessageHandling):
+class PosParser:
+    """Support class for producing Pos events."""
+
+    @classmethod
+    def _parse_xml(
```
edenhaus: Move this function to the Pos message. No need for an additional class

deebot_client/commands/xml/pos.py
```diff
@@ -16,7 +16,25 @@
...
+            p_x, p_y = p.split(",", 2)
+            p_a = xml.attrib.get("a", 0)
+            position = Position(type=position_type, x=int(p_x), y=int(p_y), a=int(p_a))
+            event_bus.notify(PositionsEvent(positions=[position]))
+            return HandlingResult.success()
+
+        return HandlingResult.analyse()
+
+
+class GetPos(XmlCommandWithMessageHandling, PosParser):
```
edenhaus:
```suggestion
class GetPos(XmlCommandWithMessageHandling, Pos):
```

deebot_client/commands/xml/pos.py
```diff
@@ -30,13 +48,21 @@ def _handle_xml(cls, event_bus: EventBus, xml: Element) -> HandlingResult:
...
-            position = Position(
-                type=PositionType.DEEBOT, x=int(p_x), y=int(p_y), a=int(p_a)
-            )
-            event_bus.notify(PositionsEvent(positions=[position]))
-            return HandlingResult.success()
+        return cls._parse_xml(PositionType.DEEBOT, event_bus, xml)
 
-        return HandlingResult.analyse()
+
+class GetChargerPos(XmlCommandWithMessageHandling, PosParser):
```
edenhaus:
```suggestion
class GetChargerPos(XmlCommandWithMessageHandling, Pos):
```

deebot_client/commands/xml/pos.py
```diff
@@ -16,7 +16,25 @@
...
-class GetPos(XmlCommandWithMessageHandling):
+class PosParser:
+    """Support class for producing Pos events."""
+
+    @classmethod
+    def _parse_xml(
+        cls, position_type: PositionType, event_bus: EventBus, xml: Element
+    ) -> HandlingResult:
+        """Handle xml message and notify the correct event subscribers."""
+        if (p := xml.attrib.get("p")) and (xml.attrib.get("valid", "1")) == "1":
```
edenhaus:
```suggestion
        if (p := xml.attrib.get("p")) and xml.attrib.get("valid") == "1":
```

We currently have tests, where always valid is present. Therefore we should analyse the message, when its missing instead assuming it's valid

edenhaus (changes requested):

deebot_client/commands/xml/pos.py
nanomad: valid=1 is missing from GetChargerPos though, I can either move the check to GetPos and Pos and remove it from the parser or assume it's "1" if the attribute is missing

edenhaus (approved):

PR #905, opened 2025-04-19
Make water amount in WaterInfoEvent optional.

nanomad:
This is needed as XML-based robots have two separate commands and events (GetWaterPermeability and GetWaterBoxInfo)

NOTE: This is a breaking change.

PR #917, opened 2025-04-22
Split WaterInfoEvent into multiple

edenhaus:
A different approach for https://github.com/DeebotUniverse/client.py/pull/905

@nanomad What do you think about this approach?


nanomad:
LGTM, I'll try and rebase my Water commands and messages on top on this and see if it works

nanomad:
@edenhaus works for me, I'll open a PR as soon as this is merged

PR #929, opened 2025-04-27
Implement XML Map messages

nanomad:


deebot_client/messages/xml/map.py
edenhaus, line 1: Can we reduce duplicated code by reusing this code in the XML map commands?

nanomad: Honestly I wouldn't know how, the logic is quite different beside extracting the "p" attribute. In the XML Map command I have to work around the limitation of _handle_xml being a class-method

edenhaus (approved):
Thanks @nanomad :+1:

PR #935, opened 2025-04-29
Fix XML Bots and add FirmwareEvent

nanomad:


nanomad:
@edenhaus Those are the bugs I found while integration testing everything:
- device.py was assuming any message could be decoded as JSON, erroring out on any XML bot message
- message.py was not decoding string payloads properly. Strange that mypy didn't catch this as we were basically ignoring some valid payload types declared by MQTT

nanomad:
@edenhaus eventually I think we should switch the "bot firmware version" to be handled differently as for XML bots we have to send a specific command (not yet implemented) so I'd consider this portion of the patch more of a workaround

deebot_client/message.py
```diff
@@ -63,6 +65,15 @@ def wrapper(
     ) -> HandlingResult:
         try:
             response = func(cls, event_bus, data)
+            # This happens if for some reason someone calls super() of an ABC where handle is not implemented
+            if not response:
+                _LOGGER.error(
+                    "Handler for message %s: %s returned no response. "
+                    "This is a bug in the implementation and should not happen",
```
edenhaus, line 72:
```suggestion
                    "This is a bug should not happen. Please report it.",
```

edenhaus:
I refactored the code to use a `FirmwareEvent`, which can be used by the XML bot in the feature too

edenhaus (approved):
Thanks @nanomad :+1:

PR #922, opened 2025-04-24
Add onStats message

gpongelli:
changes to handle onStats mqtt message 
`[deebot_client.mqtt_client] Got message: topic=iot/atr/onStats/f266fb63-5747-4cca-b8fb-3c1ceda9bd3f/659yh8/r7j2/j, payload=b'{"header":{"pri":1,"tzm":480,"ts":"1745398159530","ver":"0.0.1","fwVer":"1.4.9","hwVer":"0.1.1"},"body":{"data":{"area":27,"time":2459,"cid":"105534963","start":"1745395672","type":"auto","enablePowerMop":1,"powerMopType":2,"aiopen":1,"aitypes":[9],"avoidCount":1}}}'`

please let me know how can I test the changes on my deebot T9 AIVI, thanks.

closes #592

Issue #592, opened 2024-11-11
Deebot T9 seems not update area sensors


gpongelli:
### Checks

- [X] I have searched the existing issues and no issue is describing my issue
- [X] I have checked the [FAQ](https://deebot.readthedocs.io/home/faq/)
- [X] I have checked the [documentation](https://deebot.readthedocs.io/integrations/home-assistant)
- [X] I have installed the latest version

### The problem

Hi,
I've a Deebot T9 AIVI integrated with Home Assistant, but seems it does not update cleaned area and cleaned time spent in "real time" while cleaning is in progress.
Official Ecovacs iOS app shows that information while cleaning is in progress, so data should be available to be retrieved.

Is it possible to do so like official Ecovacs app ?

I've also post a feature request into [homeassistant community forum](https://community.home-assistant.io/t/deebot-integration-refresh-cleaned-area-while-cleaning/787660), but seems HA integration shows the content from events generated here.

if you need some trace/dump I could help but I need to be instructed to setup everything for this need.

Thanks.

### On which deebot device (vacuum) you have the issue?

Deebot T9 AIVI

### Which version of the deebot-client are you using?

8.4.0

### Country

it

### Continent

eu

edenhaus:
Yes I need logs :) Otherwise I cannot see what the bot is sending

gpongelli:
can you explain how logs can be gathered ? thanks.

edenhaus:
Enable debug logs directly on the Ecovacs config entry and attach them here

gpongelli:
Hi @edenhaus , sorry for this late reply , here it is the log file plus some screenshot
[home-assistant_ecovacs_2025-01-12T10-59-47.203Z.log](https://github.com/user-attachments/files/18389124/home-assistant_ecovacs_2025-01-12T10-59-47.203Z.log)
![IMG_2471](https://github.com/user-attachments/assets/24a1b19d-d87a-488a-a3f6-b0e339bb6abc)
![IMG_2472](https://github.com/user-attachments/assets/dc46b009-9981-46de-abcc-8b0120e80d83)
![IMG_2474](https://github.com/user-attachments/assets/d91803cb-5caf-4eec-b82f-3f9fbb1d49d8)

in the three images you can see that:

- first image -> cleaning is running at 11:48 CEST timezone (so 10:48 UTC), and it had already cleaned 6m^2 
[...]

PR #948, opened 2025-05-06
Deebot T9 AIVI correct device

gpongelli:
As suggested in https://github.com/DeebotUniverse/client.py/pull/922#pullrequestreview-2813274266 , this new PR fixes Deebot T9 AIVI device 659yh8 to correctly point to 8kwdb4 .

### DIFF
--- a/deebot_client/hardware/deebot/659yh8.py
+++ b/deebot_client/hardware/deebot/659yh8.py
@@ -1,1 +1,1 @@
-x5d34r.py
+8kwdb4.py


edenhaus:
Did you test this change?

gpongelli:
> Did you test this change?

Actually I’ve added no tests, which kind of tests are expected? 

I’ve done this change due to logs retrieved from my T9 AIVI https://github.com/DeebotUniverse/client.py/issues/592#issuecomment-2585691607

edenhaus:
Did you manually apply this change on your HA instance and checked if it is working correctly?

gpongelli:
> Did you manually apply this change on your HA instance and checked if it is working correctly?

No I didn’t. 
I based this PR’s need only on logs retrieved from vacuum.

I’ve no idea on how do test in my HA because it’s within docker container, can you point me to how to do that? Thanks.

flubshi:
@gpongelli it should work sth like that:

* Login to Home Assistant via SSH:
  * `ssh hassio@<ip>`
* Run shell in docker: 
  * `docker exec -it $(docker ps -f name=homeassistant -q) bash`
* Change to library directory:
  * `cd /usr/local/lib/python3.13/site-packages/deebot_client/hardware/deebot`
* Remove existing link:
  * `rm 659yh8.py`
* Link to 8kwdb4.py, as in this PR:
  * Link exmaple: `ln -svfT <existing_model> <your_model>`
  * `ln -svfT 8kwdb4.py 659yh8.py`
* Restart home assistant

gpongelli:
thanks @flubshi I'll try.

what I would understand is why this kind of test is needed if the two files are the same ?

edenhaus:
> what I would understand is why this kind of test is needed if the two files are the same ?

I assumed the two files are different. I now checked them and there are indeed the same. So no test are needed. Even this PR is not needed et all as the resulting file is the same so, so I will close it

gpongelli:
They refers to two different model:
* T8 is x5d34r.py
* T9 is 8kwdb4.py
* 659yh8.py is my Deebot, that is a T9 AIVI which has some video streaming more than T9

Even if files are actually the same, is more correct to point 659yh8 to 8kwdb4c, to avoid confusion if in future video part will be developed.

Issue #852, opened 2025-03-16
Ecovacs GOAT A1600 RTK Support


dcollewaert:
### Checks

- [x] I have searched the existing issues and no issue is describing my issue
- [x] I have checked the [FAQ](https://deebot.readthedocs.io/home/faq/)
- [x] I have checked the [documentation](https://deebot.readthedocs.io/integrations/home-assistant)
- [x] I have installed the latest version

### The problem

Hi,

Would it be possible to get GOAT A1600 RTK support so it could be used in Home Assistant?

### On which deebot device (vacuum) you have the issue?

GOAT A1600 RTK

### Which version of the deebot-client are you using?

2025.3.3

### Country

fr

### Continent

eu

### Anything in the logs that might be useful for us?

```txt
25-03-16 10:59:02.785 WARNING (MainThread) [deebot_client.api_client] Device class "xmp9ds" not recognized. Please add support for it: {'did': '[REMOVED]', 'name': 'E0BA34765F09HYEE0049', 'class': 'xmp9ds', 'resource': '6lZMW5q0', 'company': 'eco-ng', 'bindTs': 1741714815430, 'service': {'jmq': 'jmq-ngiot-eu.dc.ww.ecouser.net', 'mqs': 'api-ngiot.dc-eu.ww.ecouser.net'}, 'deviceName': 'GOAT A1600 RTK', 'icon': 'https://portal-ww.ecouser.net/api/pim/file/get/677cfd9e954634917988358a', 'ota': True, 'UILogicId': 'goatr_ww_h_goat2', 'materialNo': '116-2337-0000', 'pid': '65d2bb29a1bbb50225b81776', 'product_category': 'GOATBOT', 'model': 'QingGeng-R', 'updateInfo': {'needUpdate': False, 'changeLog': ''}, 'nick': 'Wall-e', 'homeId': '67cdf8f592a75c00076aa7b5', 'homeSort': 1, 'status': 1, 'btName': 'GOAT-xmp9ds-0049', 'btMac': '28:F5:2B:A5:BE:D9', 'otaUpgrade': {}}
[...]

PR #949, opened 2025-05-06
Add initial support for Goat A1600 RTK (#852)

comtel2000:
I still got issues with pytest, is there hardware required to run it?

`ERROR tests/test_device.py - NameError: name 'x5d34r' is not defined`

For now, it's just a copy of the GOAT G1 <del>without AdvancedMode and a new CustomCutMode</del>..


pyproject.toml
```diff
@@ -24,7 +24,7 @@ classifiers = [
     "Topic :: Home Automation",
     "Topic :: Software Development :: Libraries :: Python Modules",
 ]
-requires-python = ">=3.13.0"
```
edenhaus: Please revert this change

deebot_client/commands/json/custom_cut_mode.py
```diff
@@ -0,0 +1,21 @@
+"""Advanced mode command module."""
+
+from __future__ import annotations
+
+from deebot_client.events import CustomCutModeEvent
+
+from .common import GetEnableCommand, SetEnableCommand
+
+
+class GetCustomCutMode(GetEnableCommand):
```
edenhaus: What does this option do?
Can you post a screenshot from this setting in the app

edenhaus (changes requested):

deebot_client/commands/json/custom_cut_mode.py
edenhaus: Please answer this comment

comtel2000: That's what I tried to find out in my test-setup. I've tried almost everything and it's not changeable via app setting, but it probably determines which options are available in the app. It is activated for the A1600 RTK in any case.

It may enable the "Enhanced" mode..

image

pyproject.toml
comtel2000: It's reverted. For some reason it was a dependency for pytest and aiomqtt (win os)

comtel2000:
I'm just trying to get to grips with the deebot and python. The MR is only intended as a very simple basic support for the A1600.
With my very limited python skills I also like to help out with captured mitmproxy data 😄

deebot_client/events/__init__.py
```diff
@@ -308,3 +315,11 @@ class FirmwareEvent(Event):
     """Firmware event."""
 
     version: str
+
+
+@dataclass(frozen=True)
+class RainDelayEvent(Event):
+    """Rain delay event representation."""
+
+    enable: bool
```
edenhaus:
```suggestion
    enabled: bool
```

edenhaus (changes requested):
For new commands tests are required. The best would be to split adding support for the goat into a new PR, so I can merge it without waiting that all new commands are implemented. Also adding new commands need some other changes and so a split is the best imo

deebot_client/events/__init__.py
comtel2000: Ok, so I'll remove the new commands here and this MR is just a copy of the Goat G1 for the Goat R1600.

edenhaus (approved):
Thanks @comtel2000 :+1:

MilanDasek:
This is great.

One problem: Setting a moving direction is not working.

PR #960, opened 2025-05-09
Added support for T50 Max Pro (c8rj4y)

robertobasile84:
Added support for Deebot T50 max Pro Omni


edenhaus (approved):
Thanks @robertobasile84 :+1:

PR #963, opened 2025-05-12
Added support for Deebot X1 Omni China version (8onkgl)

vvd214:
Added support for Deebot X1 Omni China version (8onkgl)


edenhaus (approved):
Thanks  @vvd214 :+1:

vvd214:
Thanks

PR #984, opened 2025-05-21
Fix empty data on MapTraceEvent and log input values of rust functions on error

edenhaus:


src/map.rs
```diff
@@ -333,9 +333,13 @@ impl MapData {
...
 
     fn add_trace_points(&mut self, value: String) -> Result<(), PyErr> {
-        self.trace_points.extend(
-            extract_trace_points(value).map_err(|err| PyValueError::new_err(err.to_string()))?,
-        );
+        self.trace_points
+            .extend(extract_trace_points(&value).map_err(|err| {
+                let mut err = err.to_string();
+                err.push_str(";value:");
+                err.push_str(&value);
```
edenhaus:
```suggestion
                let err = format!("{};value:{}", err, value);
```

Code pilot missed this one

edenhaus:
pre-commit.ci autofix

PR #994, opened 2025-05-28
Add basic support for Goat A3000 (o4kvvk)

f10:
Goat A3000 is detected with this change. Also standard status for docking, battery and others is working.


knobunc:
Please add a link for cr0e4u.py too.  That is the class that my A3000 GOAT Lidar has.

edenhaus (approved):
Thanks @f10 :+1:

PR #998, opened 2025-05-29
Add support for GOAT A3000 Lidar (cr0e4u)

f10:
Change is untested as the id is different to my device.


knobunc:
FWIW I tested it and got useful results back.  I am new to the project so I can't swear they are 100% complete, but what fields are there had sane values.

edenhaus (approved):

PR #1036, opened 2025-06-20
Add support for T30C Pro (poke1m)

jjaeger94:
mapped T30C Pro to T30C Omni


edenhaus (approved):
Thanks @jjaeger94 :+1:

PR #1114, opened 2025-08-23
Add support for X9 Pro (lwmdoj)

shmuelzon:
This is a Chinese-localized version but seems to support the same features as the global variant.

Seems to work fine in HA, except for the map which, to my understanding, isn't supported on the `ilt3k8` yet either.

image


edenhaus (approved):
Thanks @shmuelzon 👍

ajroads:
Hi -thanks for working on this, when will this be available as an update for the integration?

shmuelzon:
@ajroads Looks like the library version which includes this commit (`13.7.0`) was already merged to HA (https://github.com/home-assistant/core/pull/151327) so it should be available in the next HA release, i.e., `2025.9`

ajroads:
> @ajroads Looks like the library version which includes this commit (`13.7.0`) was already merged to HA ([home-assistant/core#151327](https://github.com/home-assistant/core/pull/151327)) so it should be available in the next HA release, i.e., `2025.9`

Amazing.  Thank you!

PR #1147, opened 2025-09-06
Add support for WorkState messages/commands

shmuelzon:
Hey,

I've noticed that on my X9 Pro the status of the vacuum is always docked. This is basically set once when the robot is charged with the `GetChargeState` but never updated later. This is because the `getCleanInfo_V2` message always returns the same response regardless of the robot's status:
```json
{
    "code": 0,
    "data": {
        "state": "idle",
        "trigger": "none"
    },
    "msg": "ok"
}
```
In addition, the station state never changes as, despite it being reported as supported, it always sends the following response:
```json
{
    "code": 0,
    "data": {
        "content": {
            "error": [
                314
            ],
            "motionState": 0,
            "subContent": {
                "handEmpty": {
                    "charging": 0,
                    "connect": 0,
                    "emptying": 0,
                    "powerFull": 0
                }
            },
            "type": 0
        },
        "state": 0
    },
    "msg": "ok"
}
```

This is where the `WorkState` command/message comes in play. It includes the state of both the robot and the station together in one message, for example:
```json
{
    "data": {
        "paused": 0,
        "robotState": {
            "cleanState": {
                "cid": "122",
                "type": "freeClean"
            },
            "state": "cleaning",
            "trigger": "app"
        },
        "stationState": {
            "state": "idle",
            "trigger": "app"
        }
    }
}
```

I've added support for it but had a few questions comments before this PR will be ready for review. In no particular order:
* My robot is the X9 Pro which is linked to the X9 Pro Omni. I don't know if that one has the same issue and, as you've pointed out to me (sorry about that), the developer who added it didn't want to be disturbed. For now, I "forked" the configuration for it. Do you think that's best or should I modify the configuration for the X9 Pro Omni instead?
* I've added more states for the docking station and also changed it's enum to start from 1 like was done for others. I didn't find anywhere that might be broken from that, WDYT? I'll also add the translations for Home Assistant once this PR is merged
* I don't know what the code should do if both the robot and station is idle. In that case, I don't know if the robot is docked or not and need to rely on the `GetChargeState`. What should the `WorkState` handler do if both are idle? I'm not sure how different messages are handled if they publish the same events.
* When the robot is going back to the station for emptying before its finished its job, should its state still be "cleaning" or "returning"? For now, I switch it to "returning" only when its idle again, meaning it completed it's job
* I don't know how to force the robot into an error state so not sure how to test what it publishes in that case. Any ideas?

Thanks in advance!

### DIFF
--- a/deebot_client/events/station.py
+++ b/deebot_client/events/station.py
@@ -16,6 +16,8 @@ class State(IntEnum):
 
     IDLE = 0
     EMPTYING = 1
+    WASHING = 2
+    DRYING = 3
 
 
 @dataclass(frozen=True)
--- a/deebot_client/messages/json/work_state.py
+++ b/deebot_client/messages/json/work_state.py
@@ -1,0 +1,66 @@
+"""Base work state messages."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+from deebot_client.events import StateEvent
+from deebot_client.events.station import State as StationState, StationEvent
+from deebot_client.message import HandlingResult, MessageBodyDataDict
+from deebot_client.models import State as RobotState
+
+if TYPE_CHECKING:
+    from deebot_client.event_bus import EventBus
+
+_WORK_STATE_2_EVENTS: dict[str, dict[str, tuple[RobotState | None, StationState]]] = {
+    "idle": {
+        "idle": (None, StationState.IDLE),
+        "goCharging": (RobotState.RETURNING, StationState.IDLE),
+        "goEmptying": (RobotState.RETURNING, StationState.IDLE),
+        "emptying": (RobotState.DOCKED, StationState.EMPTYING),
+        "washing": (RobotState.DOCKED, StationState.WASHING),
+        "drying": (RobotState.DOCKED, StationState.DRYING),
+    },
+    "cleaning": {
+        "idle": (RobotState.CLEANING, StationState.IDLE),
+        "goCharging": (RobotState.RETURNING, StationState.IDLE),
+        "goEmptying": (RobotState.RETURNING, StationState.IDLE),
+        "emptying": (RobotState.DOCKED, StationState.EMPTYING),
+        "washing": (RobotState.DOCKED, StationState.WASHING),
+        "drying": (RobotState.DOCKED, StationState.DRYING),
+    },
+}
+
+
+class OnWorkState(MessageBodyDataDict):
+    """On work state message."""
+
+    NAME = "onWorkState"
+
+    @classmethod
+    def _handle_body_data_dict(
+        cls, event_bus: EventBus, data: dict[str, Any]
+    ) -> HandlingResult:
+        """Handle message->body->data and notify the correct event subscribers.
+
+        :return: A message response
+        """
+        robot_state = data.get("robotState", {}).get("state")
+        station_state = data.get("stationState", {}).get("state")
+
+        robot_status, station_status = _WORK_STATE_2_EVENTS.get(robot_state, {}).get(
+            station_state, (None, None)
+        )
+
+        if (robot_status, station_status) == (None, None):
+            return HandlingResult.analyse()
+
+        if data.get("paused") == 1:
+            robot_status = RobotState.PAUSED
+
+        if robot_status is not None:
+            event_bus.notify(StateEvent(robot_status))
+        if station_status is not None:
+            event_bus.notify(StationEvent(station_status))
+
+        return HandlingResult.success()
--- a/deebot_client/commands/json/work_state.py
+++ b/deebot_client/commands/json/work_state.py
@@ -1,0 +1,13 @@
+"""Work state commands."""
+
+from __future__ import annotations
+
+from deebot_client.messages.json.work_state import OnWorkState
+
+from .common import JsonCommandWithMessageHandling
+
+
+class GetWorkState(OnWorkState, JsonCommandWithMessageHandling):
+    """Get work state command."""
+
+    NAME = "getWorkState"
--- a/deebot_client/hardware/deebot/ilt3k8.py
+++ b/deebot_client/hardware/deebot/ilt3k8.py
@@ -38,7 +38,6 @@ from deebot_client.commands.json.child_lock import GetChildLock, SetChildLock
 from deebot_client.commands.json.clean import (
     CleanArea,
     CleanV2,
-    GetCleanInfoV2,
 )
 from deebot_client.commands.json.clean_count import GetCleanCount, SetCleanCount
 from deebot_client.commands.json.clean_logs import GetCleanLogs
@@ -54,7 +53,6 @@ from deebot_client.commands.json.life_span import GetLifeSpan, ResetLifeSpan
 from deebot_client.commands.json.network import GetNetInfo
 from deebot_client.commands.json.ota import GetOta, SetOta
 from deebot_client.commands.json.play_sound import PlaySound
-from deebot_client.commands.json.station_state import GetStationState
 from deebot_client.commands.json.stats import GetStats, GetTotalStats
 from deebot_client.commands.json.sweep_mode import GetSweepMode, SetSweepMode
 from deebot_client.commands.json.voice_assistant_state import (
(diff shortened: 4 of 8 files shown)


edenhaus:
> My robot is the X9 Pro which is linked to the X9 Pro Omni. I don't know if that one has the same issue and, as you've pointed out to me (sorry about that), the developer who added it didn't want to be disturbed. For now, I "forked" the configuration for it. Do you think that's best or should I modify the configuration for the X9 Pro Omni instead?

Normally, the same commands are supported for all X9 models, and only the PRO has more features activated. I would change the X9 capabilities and wait for users to report any issues about it. We can split them in the future if needed.

> I don't know what the code should do if both the robot and station is idle. In that case, I don't know if the robot is docked or not and need to rely on the GetChargeState. What should the WorkState handler do if both are idle? I'm not sure how different messages are handled if they publish the same events.

We can do nothing. Meaning we don't publish any event and so the current state will still be the same. I would suggest that for now and we can update or make it smarter in the future

> When the robot is going back to the station for emptying before its finished its job, should its state still be "cleaning" or "returning"? For now, I switch it to "returning" only when its idle again, meaning it completed it's job

What is the app showing?

> I don't know how to force the robot into an error state so not sure how to test what it publishes in that case. Any ideas?

You can pick up the robot and my one goes in error. But I haven't found a way to force stuck or other errors

deebot_client/events/station.py
```diff
@@ -14,8 +14,10 @@
 class State(IntEnum):
     """Enum class for all possible base station statuses."""
 
-    IDLE = 0
-    EMPTYING = 1
+    IDLE = 1
+    EMPTYING = 2
```
edenhaus: Please revert changing the current numbers, as they are mapped to Ecovacs' values. See `type` in 
https://github.com/DeebotUniverse/client.py/blob/3de30ed363385fbe7579d07ca13ef112123e0dad/deebot_client/messages/json/station_state.py#L27-L28

Even when we don't directly use these values, I want to still keep it in sync if possible

edenhaus (changes requested):
Did not review the rest of the code only answered your questions. I will draft the PR as you mention it's not ready for review. Please click "Ready on review" when it's ready for a review

shmuelzon:
> >When the robot is going back to the station for emptying before its finished its job, should its state still be "cleaning" or "returning"? For now, I switch it to "returning" only when its idle again, meaning it completed it's job

> What is the app showing?

In the app, the robot animation is the same as when it's cleaning but the text is "Empty Dust Bin". Since we don't have an "emptying" state for the robot, I think it would be better to leave it on "cleaning".
There is one issue I see with handling this message in general is that it conflicts with the `ChargeState` message since the robot will be `cleaning` while returning to the station, `docked` for a few seconds and then back to `cleaning` when the station starts emptying it. It might interfere if someone tries to automate things based on the status of the robot:
image


> > I don't know how to force the robot into an error state so not sure how to test what it publishes in that case. Any ideas?

> You can pick up the robot and my one goes in error. But I haven't found a way to force stuck or other errors

In my case the `WorkState` just says it's paused and I see a different `ErrorEvent` published with the actual problem. Here, again, there's a conflict since the state in HA will be `Error` and then, right after, `Paused`.

If there was some sort of priority mechanism where `error > cleaning > docked` and each message handler would set/unset the status flag, we could have a more consistent behavior but I'm not sure if that's something that's simple to integrate.

edenhaus:
After reading your last comment, we should not set it to cleaning when the bot returns to the station.
To avoid these state jumps, we should map the state to what the bot is doing, not if other actions are planned when the bot is docked.
So even when the bot still needs to empty the dust bin, we should set the state to returning, when it's on the way back to the station. And set it to docked, when it's actually docked. Users can have the station_state to intenfy if there are action done on the station

Sanji78:
This works properly also for my Deebot X8 OMNI PRO

shmuelzon:
Hey, I've changed the behavior to match what you requested (I think), and the PR is ready for review.

shmuelzon:
> This works properly also for my Deebot X8 OMNI PRO

@Sanji78 I'm not sure if to add it now to the X8 OMNI and/or OMNI PRO as a part of this PR or wait until it's approved and then add to other robots 🤷

Sanji78:
I will take care of updating x8 omni pro, don't worry

deebot_client/messages/json/work_state.py
```diff
@@ -0,0 +1,63 @@
...
+
+from deebot_client.events import StateEvent
+from deebot_client.events.station import State as StationState, StationEvent
+from deebot_client.message import HandlingResult, MessageBodyDataDict
+from deebot_client.models import State as RobotState
+
+if TYPE_CHECKING:
+    from deebot_client.event_bus import EventBus
+
+
```
edenhaus, line 34:
```suggestion
_WORK_STATE_2_EVENTS: dict[str, dict[str, tuple[RobotState | None, StationState]]] = {
    "idle": {
        "idle": (None, StationState.IDLE),
        "goCharging": (RobotState.RETURNING, StationState.IDLE),
        "goEmptying": (RobotState.RETURNING, StationState.IDLE),
        "emptying": (RobotState.DOCKED, StationState.EMPTYING),
        "washing": (RobotState.DOCKED, StationState.WASHING),
        "drying": (RobotState.DOCKED, StationState.DRYING),
    },
    "cleaning": {
        "idle": (RobotState.CLEANING, StationState.IDLE),
        "goCharging": (RobotState.RETURNING, StationState.IDLE),
        "goEmptying": (RobotState.RETURNING, StationState.IDLE),
        "emptying": (RobotState.DOCKED, StationState.EMPTYING),
        "washing": (RobotState.DOCKED, StationState.WASHING),
        "drying": (RobotState.DOCKED, StationState.DRYING),
    },
}


```

deebot_client/messages/json/work_state.py
```diff
@@ -0,0 +1,63 @@
...
+            ("cleaning", "idle"): (RobotState.CLEANING, StationState.IDLE),
+            ("cleaning", "goCharging"): (RobotState.RETURNING, StationState.IDLE),
+            ("cleaning", "goEmptying"): (RobotState.RETURNING, StationState.IDLE),
+            ("cleaning", "emptying"): (RobotState.DOCKED, StationState.EMPTYING),
+            ("cleaning", "washing"): (RobotState.DOCKED, StationState.WASHING),
+            ("cleaning", "drying"): (RobotState.DOCKED, StationState.DRYING),
+        }
+        robot_status, station_status = work_state_to_events.get(
+            (robot_state, station_state), (None, None)
+        )
```
edenhaus:
```suggestion
        robot_status, station_status = _WORK_STATE_2_EVENTS.get(robot_state, {}).get(
            station_state, (None, None)
        )
```

edenhaus (changes requested):
Message tests are missing.
Is it possible that one or even both states are missing in the message/command data?
If yes, we should also a test for it

deebot_client/messages/json/work_state.py
shmuelzon: Done

deebot_client/messages/json/work_state.py
shmuelzon: Done

shmuelzon:
>Message tests are missing.

Done, I needed to modify the `assert_message_failure` function a bit since the JSON messages have the firmware version in the header so that event is always published, regardless of the data that the specific handler will parse. Hope that's OK.

>Is it possible that one or even both states are missing in the message/command data?

In my logs I didn't see a case where those were missing, only the `cleanState` dictionary was included or not depending on the state of the robot. If any of the states will be missing, we'll return `analyze` so the message will appear in the logs and we can modify the handler accordingly.

edenhaus:
> In my logs I didn't see a case where those were missing, only the `cleanState` dictionary was included or not depending on the state of the robot. If any of the states will be missing, we'll return `analyze` so the message will appear in the logs and we can modify the handler accordingly.

Added tests for it to make sure we are covering them also in the future

edenhaus (approved):
Thanks @shmuelzon 👍

shmuelzon:
> Added tests for it to make sure we are covering them also in the future

Nice, now I understood what you've meant before 🤦‍♂️

> Thanks @shmuelzon 👍

And thank you for this project!

Sanji78:
I created PR 1167 to extend X8 Omni Pro
@edenhaus @shmuelzon

PR #1167, opened 2025-09-17
Add GetWorkState support for X8 Pro Omni (n0vyif)

Sanji78:
GetWorkState support for X8 Pro Omni


Sanji78 (commented):
Fixed

edenhaus (approved):
Thanks @Sanji78 👍

PR #1083, opened 2025-08-08
Add map and rooms support for T80

robin-thoni:
This adds support for the (new?) `MapInfo_V2` endpoint, used at least by the T80 robot, since this model does not support `MapSubSet`.

This PR supports:
- Basic rooms support (rooms names are not available, and are just named `Room X`)
- Map and live position support (cleaning traces are not available yet)

This has been tested on HA `2025.7.4` and `2025.8.0`: 

Image

Image

It also fixes and issue in `GetMajorMap`, where the `value` field is actually empty.

There's currently no test for that new feature. I think it's better to discuss about the general implementation first.

Also, I had to disable `pre-commit` since it was complaining about the Rust integration:
```
+ exec pylint deebot_client/map.py
building "deebot_client"
rebuilt and loaded package "deebot_client" in 2.411s
************* Module deebot_client.map
deebot_client/map.py:227:47: W0511: TODO get from MapSet_V2 (fixme)
deebot_client/map.py:12:0: E0401: Unable to import 'deebot_client.rs.map' (import-error)
deebot_client/map.py:13:0: E0401: Unable to import 'deebot_client.rs.util' (import-error)
deebot_client/map.py:30:0: E0401: Unable to import 'deebot_client.rs.map' (import-error)
```

Feel free to directly edit the PR to better match your code style, or anything.

PR #1177, opened 2025-09-20
Add support for MapInfoV2

shmuelzon:


shmuelzon:
@edenhaus I've re-implemented https://github.com/DeebotUniverse/client.py/pull/1083 and generate the map with in Rust. As I've never written in Rust before that part heavily relies on LLMs and random Internet searches so I'd appreciate any feedback there.

I'm not sure why the previous map implementation was based on a PNG (maybe the older version just sent an image?) but the new API is all XML/SVG. I tried to keep the same color scheme as the current implementation but I'm wondering if that's needed or whether I should try to keep it similar to the app, IMHO, using different colors for each room makes it a bit clearer.

I've also added support for rotating the map, as the app allows that too and I found it too odd to look at the resulting image without rotation :)

Traces and positions worked out of the box. I did notice that some traces are shown in the SVG but not in the app so I suspect there's some bug there but it also appears in the current implementation (the SVG in the test) so I don't think this commit introduced it. Maybe I'll loot at it in the future.
I've added a test for the new API.

~I did have issues with the pre-commit and the pipeline tests that I don't think are directly related to my changes but might have been triggered this the file was modified. I'd appreciate your input on those as I'm not entirely sure what should be done there. The code itself does compile and generates a map, so... 🤷‍♂️~
I still have one linter error in `tests/commands/json/test_map.py` which I've yet to solve. I'll try to get back to it but, if you have any ideas, I'd be happy to hear.

I haven't tried to integrate this to HA yet to verify it works end-to-end but hope to do so in the near future.

P.S. - just wanted to add that there's no rush on this in any way. Make sure to take care of yourself and thanks again for all you've done with this project!

Sanji78:
Following :)

Sanji78:
@shmuelzon thanks for it.

I got he following error, which I fixed:
  File "/usr/src/homeassistant/homeassistant/components/ecovacs/util.py", line 52, in get_name_key
    if enum is State.EMPTYING:
               ^^^^^^^^^^^^^^
AttributeError: type object 'State' has no attribute 'EMPTYING'

@unique
class State(IntEnum):
    """Enum class for all possible base station statuses."""

    IDLE = 0
    EMPTYING_DUSTBIN = 1
    WASHING_MOP = 2
    DRYING_MOP = 3
    EMPTYING = 4

Adding EMPTYING as state. After this fix, the map came up, but:
1) The vacuum entity doesn't have the room as attribute anymore... I would still add room_0, room_1 instead of real names... if you agree. I use them as automation to clean single rooms
2) The rooms doesn't have a colour anymore
3) Clean paths are properly displayed

image

Sanji78:
P.S. I still get those errors in the log:
2025-09-21 12:15:21.014 ERROR (MainThread) [deebot_client.message] Handler for message getMapSubSet: {'code': 20003, 'msg': 'rcp not support'} returned no response. This is a bug should not happen. Please report it.
2025-09-21 12:15:21.015 WARNING (MainThread) [deebot_client.message] Could not parse getMapSubSet: {'body': {'code': 20003, 'msg': 'rcp not support'}, 'header': {'fwVer': '1.121.0', 'hwVer': '0.1.1', 'pri': 1, 'ts': '1758449720071', 'tzm': 480, 'ver': '0.0.1', 'wkVer': '0.1.54'}}
2025-09-21 12:15:21.015 WARNING (MainThread) [deebot_client.message] Could not parse getMapSubSet: {'body': {'code': 20003, 'msg': 'rcp not support'}, 'header': {'fwVer': '1.121.0', 'hwVer': '0.1.1', 'pri': 1, 'ts': '1758449720071', 'tzm': 480, 'ver': '0.0.1', 'wkVer': '0.1.54'}}
2025-09-21 12:15:21.015 WARNING (MainThread) [deebot_client.command] Could not parse getMapSubSet for n0vyif: {'ret': 'ok', 'resp': {'body': {'code': 20003, 'msg': 'rcp not support'}, 'header': {'fwVer': '1.121.0', 'hwVer': '0.1.1', 'pri': 1, 'ts': '1758449720071', 'tzm': 480, 'ver': '0.0.1', 'wkVer': '0.1.54'}}, 'id': 'Yipl', 'payloadType': 'j'}
2025-09-21 12:15:21.022 ERROR (MainThread) [deebot_client.message] Handler for message getMapSubSet: {'code': 20003, 'msg': 'rcp not support'} returned no response. This is a bug should not happen. Please report it.
2025-09-21 12:15:21.022 WARNING (MainThread) [deebot_client.message] Could not parse getMapSubSet: {'body': {'code': 20003, 'msg': 'rcp not support'}, 'header': {'fwVer': '1.121.0', 'hwVer': '0.1.1', 'pri': 1, 'ts': '1758449720090', 'tzm': 480, 'ver': '0.0.1', 'wkVer': '0.1.54'}}
2025-09-21 12:15:21.023 WARNING (MainThread) [deebot_client.message] Could not parse getMapSubSet: {'body': {'code': 20003, 'msg': 'rcp not support'}, 'header': {'fwVer': '1.121.0', 'hwVer': '0.1.1', 'pri': 1, 'ts': '1758449720090', 'tzm': 480, 'ver': '0.0.1', 'wkVer': '0.1.54'}}
2025-09-21 12:15:21.023 WARNING (MainThread) [deebot_client.command] Could not parse getMapSubSet for n0vyif: {'ret': 'ok', 'resp': {'body': {'code': 20003, 'msg': 'rcp not support'}, 'header': {'fwVer': '1.121.0', 'hwVer': '0.1.1', 'pri': 1, 'ts': '1758449720090', 'tzm': 480, 'ver': '0.0.1', 'wkVer': '0.1.54'}}, 'id': 'JWRS', 'payloadType': 'j'}
2025-09-21 12:15:22.331 ERROR (MainThread) [deebot_client.message] Handler for message getCleanPreference: {'code': 20005, 'msg': 'call setting service empty'} returned no response. This is a bug should not happen. Please report it.
2025-09-21 12:15:22.331 WARNING (MainThread) [deebot_client.message] Could not parse getCleanPreference: {'body': {'code': 20005, 'msg': 'call setting service empty'}, 'header': {'fwVer': '1.121.0', 'hwVer': '0.1.1', 'pri': 1, 'ts': '1758449721396', 'tzm': 480, 'ver': '0.0.1', 'wkVer': '0.1.54'}}
2025-09-21 12:15:22.331 WARNING (MainThread) [deebot_client.message] Could not parse getCleanPreference: {'body': {'code': 20005, 'msg': 'call setting service empty'}, 'header': {'fwVer': '1.121.0', 'hwVer': '0.1.1', 'pri': 1, 'ts': '1758449721396', 'tzm': 480, 'ver': '0.0.1', 'wkVer': '0.1.54'}}
2025-09-21 12:15:22.331 WARNING (MainThread) [deebot_client.command] Could not parse getCleanPreference for n0vyif: {'ret': 'ok', 'resp': {'body': {'code': 20005, 'msg': 'call setting service empty'}, 'header': {'fwVer': '1.121.0', 'hwVer': '0.1.1', 'pri': 1, 'ts': '1758449721396', 'tzm': 480, 'ver': '0.0.1', 'wkVer': '0.1.54'}}, 'id': 'zPqZ', 'payloadType': 'j'}

Sanji78:
@shmuelzon If you are interested, I created a js file with which you can selectively remove backgroud colour, walls, fornitures, cleaning paths... the idea is to overlap this map to the floorplan for example :)

It basically takes the image generated by the image entity, retrieve the original SVG and patch it.

I don't know where and how to include it in your PR.

image

Sanji78:
This is the code to add the rooms. If you give me access to the PR I can patch it.
@shmuelzon 

on_map_info inside class MapV2:

```
        async def on_map_info(event: MapInfoEvent) -> None:
            self._map_data.set_map_info(event.info)
            rooms = {}
            for room in event.coordinates_map[0][1:]:
                subset_id = room.split(";")[0]
                room_coordinates = room[len(subset_id) + 1 :]
                room_name = (
                    self._rooms[int(subset_id)].name
                    if int(subset_id) in self._rooms
                    else f"Room {subset_id}"  # TODO get from MapSet_V2
                )
                rooms[int(subset_id)] = Room(
                    room_name,
                    int(subset_id),
                    room_coordinates,
                )
            self._rooms = rooms
            def parse_coordinates(
                coordinates: str,
            ) -> tuple[int, list[dict[str, int]]]:
                """Parse coordinates from a string."""
                coords_split = coordinates.strip().split(";")
                room_id = int(coords_split[0])
                coords_str = coords_split[1:] if len(coords_split) > 1 else []
                coords = []
                for coord in coords_str:
                    c_s = coord.split(",")
                    coords.append(
                        {
                            "x": int(c_s[0]),
                            "y": int(c_s[1]),
                        }
                    )
                return (room_id, coords)

            self._map_info_v2 = {}
            for type in event.coordinates_map:
                self._map_info_v2[int(type[0])] = {}
                for room_data in type[1:]:
                    room_id, room_coords = parse_coordinates(room_data)
                    self._map_info_v2[int(type[0])][room_id] = room_coords

            self._event_bus.notify(RoomsEvent(list(self._rooms.values())))
            self._event_bus.notify(MapChangedEvent(datetime.now(UTC)), debounce_time=1)
```

_handle_body_data_dict of class GetMapInfoV2 inside json/map:

```
    @classmethod
    def _handle_body_data_dict(
        cls, event_bus: EventBus, data: dict[str, Any]
    ) -> HandlingResult:

        coordinates = decompress_base64_data(data["info"]).decode()
        coordinates_map = json.loads(coordinates)
        
        event_bus.notify(
            MapInfoEvent(
                map_id=data["mid"],
                info=data["info"],
                coordinates_map=coordinates_map,
            )
        )
```

MapInfoEvent inside events/map:
```

@dataclass(frozen=True)
class MapInfoEvent(Event):
    """Map info event representation."""

    map_id: str
    info: str
    coordinates_map: list[list[str]]
```

shmuelzon:
@Sanji78 

> I got he following error, which I fixed:

That error is coming from https://github.com/DeebotUniverse/client.py/commit/91713302a62f2ba6b9145cbb629bc97792b8bd7b which, to my understanding, should be addressed now in HA with something like:
```patch
diff --git a/homeassistant/components/ecovacs/util.py b/homeassistant/components/ecovacs/util.py
index 968ab92851b..fb98ea7a1cf 100644
--- a/homeassistant/components/ecovacs/util.py
+++ b/homeassistant/components/ecovacs/util.py
@@ -49,9 +49,6 @@ def get_supported_entities(
 @callback
 def get_name_key(enum: Enum) -> str:
     """Return the lower case name of the enum."""
-    if enum is State.EMPTYING:
-        # Will be fixed in the next major release of deebot-client
-        return "emptying_dustbin"
     return enum.name.lower()


```
I believe it will be part of https://github.com/home-assistant/core/pull/152448/files when the new release of this library is ready.

> The vacuum entity doesn't have the room as attribute anymore... I would still add room_0, room_1 instead of real names... if you agree. I use them as automation to clean single rooms

I plan to address the rooms in a following commit. I see the information available in the `onSimpleARMapSet` message and `MapSet_V2`. I'll probably use the latter as it already exists in this library but I'm not sure why the information isn't used from there to publish the event with the rooms information but rather it's used only to detect rooms that were removed and the `MapSubSet` is used to publish them. Anyway, I don't want to complicate this commit more than it already is.

> The rooms doesn't have a colour anymore

Yes, this is on purpose to match the current look and feel of the maps. I asked edenhaus about this in the comment above and will wait for his feedback as to whether we should keep a unified UI or if we want it to look similarly to the app.

> P.S. I still get those errors in the log:
2025-09-21 12:15:21.014 ERROR (MainThread) [deebot_client.message] Handler for message getMapSubSet: {'code': 20003, 'msg': 'rcp not support'} returned no response. This is a bug should not happen. Please report it.

I'm not sure where this comes from need to see who requests this API.

>  I created a js file with which you can selectively remove backgroud colour, walls, fornitures, cleaning paths... the idea is to overlap this map to the floorplan for example :)

That's a nice idea! Not sure it should be part of this repository but 🤷‍♂️ . In any case, I don't have any way to use it as my floorplan is 3D so just mapping the coordinates will be a nightmare. Shameless plug: https://github.com/shmuelzon/home-assistant-floor-plan/

> This is the code to add the rooms. If you give me access to the PR I can patch it.

See above.

edenhaus:
Will review the code later but first I will answer your comment:

> I'm not sure why the previous map implementation was based on a PNG (maybe the older version just sent an image?)

I did not even create the first map, and PNG was chosen. Some time ago, we migrated some parts of it to SVG, but the bot sends a two-dimensional array about the structure, where, for example, 1=wall, 2=floor, 3=carpet (only guessing the values), and it was easier to create a PNG out of it. Feel free to look into it, if we can also convert that one into a SVG :)

>  I tried to keep the same color scheme as the current implementation but I'm wondering if that's needed or whether I should try to keep it similar to the app, IMHO, using different colors for each room makes it a bit clearer.

The Ecovacs app combines multiple apps in it. If you ever need to reverse engineer it for some commands or handling, you will see it. So keeping the colors similar to the app is impossible as different models have different colors, but we can add colorized rooms in a follow-up PR (to keep this one as small as possible).

> I still have one linter error in tests/commands/json/test_map.py which I've yet to solve. I'll try to get back to it but, if you have any ideas, I'd be happy to hear.

I will look at it during the review :)

> P.S. - just wanted to add that there's no rush on this in any way. Make sure to take care of yourself and thanks again for all you've done with this project!

Thanks a lot :heart: 

> I believe it will be part of [home-assistant/core#152448 (files)](https://github.com/home-assistant/core/pull/152448/files) when the new release of this library is ready.

Yes, that's correct. I will update the HA code during the version bump.

Please don't use any more force pushes as it makes it difficult to review, as otherwise I can see which changes you have made between my reviews. A squash commit is created anyway on merge, so it makes no difference how many commits the PR has.

deebot_client/commands/json/map/__init__.py
```diff
@@ -393,3 +400,24 @@ def _handle_body_data_dict(
...
         return HandlingResult.analyse()
+
+
+class GetMapInfoV2(JsonCommandWithMessageHandling, MessageBodyDataDict):
+    """Get map info v2 command."""
+
+    NAME = "getMapInfo_V2"
+
+    def __init__(self, map_id: str = "") -> None:
+        super().__init__({"mid": map_id, "type": "0"})
```
edenhaus, line 388: From the other PR:
Do you know what type 0 stands for?
Does other types also exist?

deebot_client/events/map.py
```diff
@@ -95,6 +103,7 @@ class CachedMapInfoEvent(Event):
 
     name: str
     active: bool = field(kw_only=True)
+    angle: int = field(default=0, kw_only=True)
```
edenhaus:
```suggestion
    angle: int = field(kw_only=True)
```

deebot_client/map.py
```diff
@@ -102,6 +103,7 @@ async def on_major_map(event: MajorMapEvent) -> None:
                     if (
                         self._map_data.map_piece_crc32_indicates_update(idx, value)
                         and event.requested
+                        and self._capabilities.minor is not None
```
edenhaus: Capabilities are fixed and should not checked multiple times in the loop.
Please out of the TaskGroup

```python
if self._capabilities.minor is None:
    return
```

deebot_client/map.py
```diff
@@ -184,6 +186,87 @@ async def teardown(self) -> None:
         self._unsubscribers.clear()
 
 
+class MapV2:
```
edenhaus: Do we really need this V2 object if all the code is in rust?

src/map.rs
```diff
@@ -105,9 +110,9 @@ fn points_to_svg_path(points: &[Point]) -> Option<String> {
         }
 
         if !p.connected {
-            let space = if 0.0 < y { " " } else { "" };
-            let _ = write!(svg_path, "m{x}{space}{y}");
-            last_command = SvgPathCommand::MoveBy;
+            let space = if 0.0 < p.y { " " } else { "" };
+            let _ = write!(svg_path, "M{}{}{}", p.x, space, p.y);
+            last_command = SvgPathCommand::MoveTo;
```
edenhaus: Why did you change this?

src/map.rs
```diff
@@ -293,6 +312,7 @@ struct Position {
     y: i32,
 }
 
+#[cfg(test)]
```
edenhaus:
```suggestion
```

src/map.rs
```diff
@@ -575,8 +730,89 @@ impl MapData {
...
+                _ => {}
+            }
+        }
+        self.outlines = outlines;
+        self.areas = areas;
+        self.block_lines = block_lines;
+    }
 }
 
+#[cfg(test)]
```
edenhaus:
```suggestion
```

edenhaus (changes requested):
I just did a quick code review. I need to still look into detail about the rust code but in the meantime here a quick feedback

deebot_client/commands/json/map/__init__.py
shmuelzon: No idea what it means and I haven't seen a different value. I just see that on the MQTT bus in the command when the app is open.

deebot_client/events/map.py
shmuelzon: I gave it a default value so I won't need to change any other code that publishes this event and doesn't have the angle. Why do you suggest removing it?

deebot_client/map.py
shmuelzon: I think that we listen to different events and also handle them differently. For example, the rooms will probably need to be handled differently as the "ar" message tells us their names (where in the current implementation it's only used to detect rooms that were removed) and the `MapInfo_V2` message can be used to get the room coordinates, which, BTW, I'm not sure who even uses.

src/map.rs
shmuelzon: This was a refactor done by the LLM and it commented it as a "more stable" version. Technically speaking, it makes more sense to me to "jump" to the next coordinate instead of doing a relative (calculated) move. I also thought that it might be redundant but the map didn't turn out properly when I reverted it back to the original `MoveBy` version so I kept it for now.

src/map.rs
shmuelzon: I think I needed to add this for the Rust linter to pass, are you sure it's not needed?

shmuelzon:
> Please don't use any more force pushes as it makes it difficult to review, as otherwise I can see which changes you have made between my reviews. A squash commit is created anyway on merge, so it makes no difference how many commits the PR has.

I, sometimes, want multiple commits in a single pull request in case they're all needed to support a feature but each has a different logical change so I'm accustomed to amending commits. But sure, I'll do that going forward here.

BTW - What's the best way to install a custom version of this library on my HA docker to test it end-to-end? Up until now I've only made changes to Python files so I just patched the files in the running Docker but now I also need the binary (musl-based) library. Any tips?

src/map.rs
edenhaus: This annotation says this is a test config, which is for sure wrong as this function was used in the code before

src/map.rs
edenhaus: If the code is not interchangeable, we need to see if we can change it because older maps are using it.

Can you please verify that the code works correctly? I just viewed the SVG specified in test_2.py, and it does not look correct. 
image

edenhaus:
> BTW - What's the best way to install a custom version of this library on my HA docker to test it end-to-end? Up until now I've only made changes to Python files so I just patched the files in the running Docker but now I also need the binary (musl-based) library. Any tips?

If the CI is successfully it creates the wheels for you :)
image
- Just go on the summary of the CI workflow and scroll down to the artifacts.
- Download the one for your system
- Unzip the file
- Install it with `uv pip install deebot_client-*` Assuming you have only one wheel in the current folder otherwise you need to specify the exact filename.
- Wheel and zip file can be deleted

P.S. Obviously you need to install the wheel inside the container

src/map.rs
```diff
@@ -423,36 +464,91 @@ impl MapData {
...
         }
+
         if let Some(trace) = get_trace_path(&self.trace_points) {
-            document.append(trace);
+            map_layer.append(trace);
         }
-        for position in get_svg_positions(&positions, &viewbox) {
-            document.append(position);
+        if let Some(vb) = viewbox_opt.as_ref() {
+            for position in get_svg_positions_rotated(&positions, vb, self.rotation_deg) {
```
edenhaus: Why are you rotating some parts directly and some (map_layer) via svg transform?

src/map.rs
edenhaus: These functions are no longer used as you use the rotated one. Therefore the linter complains about an unused function

deebot_client/events/map.py
edenhaus: It would be only another place in the code but we could keep the default value

deebot_client/map.py
edenhaus: Done it for you

deebot_client/map.py
edenhaus: Done it for you. One Map object simplyfies the logic

src/map.rs
shmuelzon: OK, I understand now what the problem was with the relative move and it was simply a bug. In my case, the relative move on the y axis was 0 while the X was something, e.g., 40. The code would then generate a relative move of 140 on the x axis because it didn't include the space between them. So I reverted back to the relative move and fixed that bug.
As for the resulting cropped image, this is because the SVG viewbox didn't take into consideration the rotation of the image, so I added that too and I think it's OK now.

src/map.rs
shmuelzon: Right, I cleaned those functions up a bit

src/map.rs
shmuelzon: Same, did some cleanups

src/map.rs
shmuelzon: My issue was that the "icons" for the robot and it's station were also rotated along with the rest of the map. It's not a problem for the robot since it's just a circle but the pin for the station would be rotated. Because of that the "position" icons are added outside of the map rotation but I still need to calculate the correct coordinates to put them on the rotated map. We could, perhaps, counter-rotate the "positions" icons and then place them on the same layer as the map. It'll remove the need to calculate the rotated location but, 🤷‍♂️

edenhaus: It's only strange to have some parts rotated later in the browser and some during generation...
I prefer one way.

src/map.rs
```diff
@@ -575,23 +760,122 @@ impl MapData {
...
+fn get_svg_positions<'a>(
+    positions: &'a [Position],
+    viewbox: &ViewBox,
+    rotation_deg: f32,
+) -> Vec<Use> {
     let mut positions: Vec<&'a Position> = positions.iter().collect();
     positions.sort_by_key(|d| d.position_type.order());
-    debug!("Adding positions: {positions:?}");
+
+    let theta = rotation_deg.to_radians();
```
edenhaus: There is duplicated code in this PR, which should be avoided.
Please create a function that is called in multiple places. This will help make the code more readable.

src/map.rs
```diff
@@ -423,36 +455,91 @@ impl MapData {
...
+
+        // Create layers: map_layer (rotated), ui_layer (not rotated)
+        let mut map_layer = Group::new();
+        let mut ui_layer = Group::new();
+        if let Some(vb) = viewbox_opt.as_ref() {
+            if self.rotation_deg != 0.0 {
+                let cx = vb.min_x as f32 + (vb.width as f32 / 2.0);
+                let cy = vb.min_y as f32 + (vb.height as f32 / 2.0);
+                map_layer = map_layer.set(
+                    "transform",
```
edenhaus: You are already calculating the viewbox and some parts with rotation in mind. We should calculate everything instead of using transform here, as otherwise the code is strange

edenhaus (changes requested):

edenhaus:
Code review by Claude AI, which I copied and pasted directly here, as for some reason, it can review a fork (or I did something wrong). I know you don't are familiar with Rust and that's also not a problem. Try to fix the issues and if you are struggling if one, please ask me for help.

# Major Issues Found:
## 1. Code Duplication and Poor Structure

This helper function is doing too much and is poorly structured
```rust
fn add_polygons_to_group(mut group: Group, polygons: &Vec<Vec<Point>>, fill_color: &str) -> Group {
    for polygon in polygons {
        if polygon.len() >= 3 {
            let mut coords: Vec<f32> = Vec::with_capacity(polygon.len() * 2);
            for point in polygon {
                coords.push(point.x);
                coords.push(point.y);
            }
            group = group.add(Polygon::new().set("fill", fill_color).set("points", coords));
        }
    }
    group
}
```

### Problems:

- Takes `&Vec<Vec<Point>>` instead of `&[Vec<Point>]` (anti-pattern)
- Logic is repeated in multiple places for different polygon types
- Could be more generic

## 2. Complex, Hard-to-Follow Logic

The to_svg method became a massive function with complex nested logic:
This entire section is confusing and hard to follow
```rust
let (viewbox_opt, bg_img): (Option<ViewBox>, Option<String>) =
    if let Some((base64_image, viewbox)) = background {
        document = document.set("viewBox", viewbox.to_svg_viewbox(self.rotation_deg));
        (Some(viewbox), Some(base64_image))
    } else if !self.outlines.is_empty()
        || !self.areas.is_empty()
        || !self.block_lines.is_empty()
    {
        if let Some(vb) =
            viewbox_from_shapes(&self.outlines, &self.areas, &self.block_lines)
        {
            document = document.set("viewBox", vb.to_svg_viewbox(self.rotation_deg));
            (Some(vb), None)
        } else {
            return Ok(None);
        }
    } else {
        return Ok(None);
    };
```

## 3. Poor Error Handling

```rust
let _segment_id = parts.next(); // Ignoring values with underscore prefix
```

This pattern appears multiple times - the author is ignoring potential parsing errors.

## 4. Inefficient String/Data Processing

Inefficient parsing in parse_map_info
```rust
let mut it = spec.splitn(3, ',');
let x = it.next().and_then(|v| v.parse::<i32>().ok());
let y = it.next().and_then(|v| v.parse::<i32>().ok());
let t = it.next().unwrap_or("1").trim().to_string(); // Unnecessary allocation
```

## 5. Complex Mathematical Logic Without Comments

Rotation calculation is hard to understand and is used slightly differently throughout the code
```rust
let theta = rotation_deg.to_radians();
let cos_t = theta.cos();
let sin_t = theta.sin();
let cx = viewbox.min_x as f32 + (viewbox.width as f32 / 2.0);
let cy = viewbox.min_y as f32 + (viewbox.height as f32 / 2.0);
// ... more complex math without explanation
```

## 6. Inconsistent Data Types

Mixed usage of f32, i32, i16, u16 without clear reasoning for the conversions.

## 7. Code Duplication: 

The parse_map_info method has three nearly identical parsing loops for outlines (type "1"), areas (type "2"), and block_lines (type "6"). This should be abstracted into helper functions.

## 8. Non-idiomatic Rust:

- Manual string splitting and parsing instead of using proper parsers
- Unnecessary .clone() calls and string allocations
- Complex nested logic that could use iterators and functional patterns
- Poor error handling (silent failures on parse errors)

## 9. Messy Structure:

- Visual constants added at top without grouping/documentation
- Complex rotation logic mixed into viewbox calculations
- The get_svg_positions function became overly complex with rotation math inline

## 10. Removed Tests: 

The calc_point_in_viewbox function and its tests were removed, but similar logic was duplicated elsewhere.

## 11. Poor Separation of Concerns:

Map rendering, coordinate transformation, and data parsing are all mixed together.

src/map.rs
shmuelzon: Would you prefer not using SVG rotation at all then and calculate the rotated points always? This also means the image used in the current map implementation will also need to be "manually" rotated. Perhaps it would be best to leave the rotation out of this pull request and leave it as a following feature?

src/map.rs
```diff
@@ -105,7 +111,7 @@ fn points_to_svg_path(points: &[Point]) -> Option<String> {
         }
 
         if !p.connected {
-            let space = if 0.0 < y { " " } else { "" };
+            let space = if 0.0 <= y { " " } else { "" };
```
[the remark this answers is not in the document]
shmuelzon: It is intentional and was previously a bug

shmuelzon:
> Code review by Claude AI, which I copied and pasted directly here, as for some reason, it can review a fork (or I did something wrong). I know you don't are familiar with Rust and that's also not a problem. Try to fix the issues and if you are struggling if one, please ask me for help.

@edenhaus Is this CR from the latest version? I think some of those were already changed.

In any case, I removed the map rotation code for now and refactored the remaining code. Hope it's better now.

src/map.rs
shmuelzon: Removed rotation from this PR

src/map.rs
shmuelzon: Removed rotation from this PR

tests/test_map.py
```diff
@@ -201,7 +204,7 @@ async def test_empty_maptrace(
         map_obj = await setup_map(execute_mock, event_bus, static_device_info)
         event_bus.notify(MapTraceEvent(0, 0, ""))
         await block_till_done(event_bus)
-        map_obj._map_data.add_trace_points.assert_not_called()
+        cast("Mock", map_obj._map_data.add_trace_points).assert_not_called()
```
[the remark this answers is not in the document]
shmuelzon: Ruff add those quotes, so I'm not sure what to do with this...

deebot_client/capabilities.py
```diff
@@ -177,8 +177,11 @@ class CapabilityMap:
     cached_info: CapabilityEvent[CachedMapInfoEvent]
     changed: CapabilityEvent[MapChangedEvent]
     clear: CapabilityExecute[[]] | None = None
-    major: CapabilityEvent[MajorMapEvent] | CapabilitySet[MajorMapEvent, [str]]
-    minor: CapabilityExecute[[int, str]]
+    info: CapabilityExecute[[str]] | None = None
+    major: (
```
edenhaus: `setMajorMap` is used to select the map, which you want to use. The bot can store up to 4 maps/floorplans.
How are new bots changing the map?

tests/commands/json/map/test_cached_map_info.py
```diff
@@ -13,16 +16,22 @@
...
 
 @pytest.mark.parametrize(
-    ("device_class", "map_set_type"),
+    ("device_class", "map_set_type", "additional_commands"),
     [
-        ("yna5xi", GetMapSet),
-        ("kr0277", GetMapSetV2),
+        ("yna5xi", GetMapSet, []),
+        ("kr0277", GetMapSetV2, []),
+        ("lwmdoj", GetMapSetV2, [GetMapInfoV2]),
```
edenhaus:
```suggestion
    ("device_class", "map_set_type", "map_info_command"),
    [
        ("yna5xi", GetMapSet, None),
        ("kr0277", GetMapSetV2, None),
        ("lwmdoj", GetMapSetV2, GetMapInfoV2),
```

The list will only make things complex and also the first argument needs to be the map_id

tests/commands/json/map/test_cached_map_info.py
```diff
@@ -13,16 +16,22 @@
...
     ],
 )
 async def test_getCachedMapInfo(
-    device_class: str, map_set_type: type[GetMapSet | GetMapSetV2]
+    device_class: str,
+    map_set_type: type[GetMapSet | GetMapSetV2],
+    additional_commands: list[type[Command]],
```
edenhaus:
```suggestion
    map_info_command: Callable[[str],Command] | None,
```

tests/commands/json/map/test_init.py
```diff
@@ -488,3 +489,24 @@ async def test_getMapTrace() -> None:
...
     )
+
+
+async def test_getMapInfoV2() -> None:
+    mid = "98100521"
+    info = "KLUv/QRYmQAAW1siMSJdLFsiMiJdLFsiNiJdXbBRuA4="
+    (
+        json,
+        firmware_event,
+    ) = get_request_json(
```
edenhaus:
```suggestion
    (json, firmware_event) = get_request_json(
```

src/map.rs
```diff
@@ -575,6 +659,99 @@ impl MapData {
...
+        };
+
+        let mut outlines = Vec::new();
+        let mut areas = Vec::new();
+        let mut block_lines = Vec::new();
+
+        for group in &info {
+            let Some(first) = group.first() else { continue };
+            match first.as_str() {
+                "1" => {
```
edenhaus: Please create constant for it with a name that describes it

src/map.rs
```diff
@@ -575,6 +659,99 @@ impl MapData {
...
+                        }
+
+                        if let Some(first) = path_points.first().filter(|p| p.connected) {
+                            path_points.push(Point {
+                                x: first.x,
+                                y: first.y,
+                                connected: true,
+                            });
+                        }
+                        outlines.push(path_points);
```
edenhaus: Move it into a separate function

src/map.rs
```diff
@@ -575,6 +659,99 @@ impl MapData {
...
+                    }
+                }
+                "2" | "6" => {
```
edenhaus: Same here

src/map.rs
```diff
@@ -575,6 +659,99 @@ impl MapData {
...
+                            .filter(|s| !s.is_empty())
+                            .skip(1)
+                            .filter_map(parse_coords)
+                            .map(|(x, y)| calc_point(x, y))
+                            .collect();
+
+                        if poly_points.len() >= 3 {
+                            dest.push(poly_points);
+                        }
+                    }
```
edenhaus: Move it into a separate function, which should return dest so we can assign it to the correct value instead during the magic above

src/map.rs
```diff
@@ -575,6 +659,99 @@ impl MapData {
             view_box,
         )))
     }
+
+    fn parse_map_info(&mut self, info: MapV2Info) {
+        let parse_coords = |s: &str| -> Option<(f32, f32)> {
```
edenhaus: Separate function

src/map.rs
```diff
@@ -575,6 +659,99 @@ impl MapData {
...
+                        &mut areas
+                    } else {
+                        &mut block_lines
+                    };
+
+                    for entry in group.iter().skip(1).filter(|e| !e.is_empty()) {
+                        let poly_points: Vec<Point> = entry
+                            .split(';')
+                            .filter(|s| !s.is_empty())
+                            .skip(1)
```
edenhaus: Please add comments. For example why are we skipping here

src/map.rs
```diff
@@ -575,6 +659,99 @@ impl MapData {
...
+        }
+
+        self.outlines = outlines;
+        self.areas = areas;
+        self.block_lines = block_lines;
+    }
+
+    fn viewbox_from_outlines(&self) -> Option<ViewBox> {
+        let mut bounds = None;
+        self.outlines
```
edenhaus: You are iterating over the outlines multiple times during the map generation. Can we reduce it?

edenhaus (changes requested):

deebot_client/capabilities.py
shmuelzon: Not sure, is this the multi-floor mode thing? I need to see if I can enable that and see what happens when my apartment is clear for it to map the "second" floor. In any case, even when the app is open, I don't see any logs with `major|minor` in them so I'm not sure it's even supported, at least not with my robot.

Will update....

tests/commands/json/map/test_cached_map_info.py
shmuelzon: Removed the `additional_commands` and now rely on the capabilities for whether `GetMapInfoV2` is expected or not. Hope that's better.

If I'm already loading the capabilities of the device, should I also get from there whether `GetMapSet` or `GetMapSetV2` is expected?

tests/commands/json/map/test_cached_map_info.py
shmuelzon: Removed `additional_commands` altogether

tests/commands/json/map/test_init.py
shmuelzon: Done, without the parentheses

src/map.rs
shmuelzon: Done

src/map.rs
shmuelzon: Done

src/map.rs
shmuelzon: Done

src/map.rs
shmuelzon: Assuming I've understood correctly, done.

I'm wondering if the code should continue assuming that each type (outline, area and block lines) might appear multiple types and extend the vector each time or if we can simplify and just assigned the parse points directly to `self.outlines`, etc.

src/map.rs
shmuelzon: Done

src/map.rs
shmuelzon: Done

src/map.rs
shmuelzon: The only way I can think of to make this happen is if the `points_to_svg_path` will return both that path and the ViewBox, similarly to how `generate_background_image` does it. It does mean that for any other calls to `points_to_svg_path` we'll, needlessly, calculate the box but that might not be too computationally heavy. Would you like me to try and do that?

deebot_client/capabilities.py
shmuelzon: Hey, I looked into this today and here's what I got:
The `setMajorMap` does seem to work also for the newer robots, however, the mobile app uses a different command to change the active map, `setCachedMapInfo`, with the following payload:
```json
{"body":{"data":{"act":"userchangemap","userUsing":1,"mid":"379107632"}}}
```
As an aside, I think this one is used to change the name of the map:
```json
{"body":{"data":{"act":"mod","mid":"379107632","name":"Map 2"}}}
```
So, for now, I guess I'll revert the removal of major/minor from the robots' configuration but, perhaps, in the future, it should be added/changed to the new API.

Another thing I noticed, which I'm not sure how it should work, is that the map image isn't updated to reflect the newly select map. I do see the `onCachedMapInfo` message being sent a few times (some without any map marked as `using`, and then others with the new map as `using`) but this doesn't trigger the `getMapInfo_V2` command in order to get the new layout. Currently it's only called when handling the response to `getCachedMapInfo` but I don't understand how it should be handled when `onCachedMapInfo` is received. That handler publishes an `CachedMapInfoEvent` that doesn't do anything in `deebot_client/map.py`. Should the `onCachedMapInfo` also trigger somehow the response handler for `getCachedMapInfo` since that handler also requests the other map information for rooms, virtual walls, etc.

@edenhaus Please let me know what you think regarding bringing back the major/minor configuration to the robots and about fetching the new map info when a new map is active.

tests/commands/json/map/test_cached_map_info.py
```diff
@@ -85,12 +102,12 @@ async def test_getCachedMapInfo(
                     ),
                 }
             ),
-            *[firmware_event for _ in MapSetType],
+            *[firmware_event for _ in range(len(expected_commands))],
```
lcnittl:
```suggestion
            *[firmware_event for _ in expected_commands],
```
this should do the same without overhead for calculating length and range thereof

thanks for the PR!

edenhaus:
I have refactored map.rs on dev into submodules to keep it small and manageable.
i suggest we move the new code also in a submodule `map_info` or something similar. Feel free to do it otherwise I will give it a try next week

shmuelzon:
> I have refactored map.rs on dev into submodules to keep it small and manageable. i suggest we move the new code also in a submodule `map_info` or something similar. Feel free to do it otherwise I will give it a try next week

@edenhaus, I gave it a go. Let me know what you think...

edenhaus:
I will look at it next week :)

As I analyzed my bot today, I just wanted to ask out of curiosity if your bot also supports the `getMapInfo` command.
In particular, I analyzed all my `atr` messages and saw that my bot is also sending `onMapInfo` and also `onMapInfo_V2`

Below is the data structure of the `body` of a `onMapInfo_V2`:

```json
{
    "data": {
        "mid": "1132127808",
        "batid": "idfhjd",
        "serial": "1",
        "index": "1",
        "type": "1,2,5",
        "outlineVer": "1",
        "info": "XQAABAACFgAAAC2WwEIAXhHX5+ZnxrU6fNSvxg2D8m6mb4wWV5FH7sAH5xeaoDuPMNKyNe9eOEIKXaOvBD3Zx3p7+yetAOcp6MpoJkhfK7ZNs2NzxNjooHtxZuxIokSd0TxcRJHWkxhksTNMlo++jpReOLCDJo8dBFoIrdlNZX7GsqIR2tMc1brGTSJhglu1MeidPrpIxN4T5juxJNAvaFkHFLs60DgNfbRqoq25eq620UtlIJF6bKvi71aYlwIoLGMRCYyHF9zvB7LfB59/CsmuCXIf946MaCzzxBnkuIwHxXPFlhT/YkMtGSpaDCY5suez7FllnXbbb5EsC2GZXwfTUKBpQZcautb+r12hXkQZ7uibO0KDLPW1gt5gQGVPK3qiWJegVWOsMSG0Efr/mnUTzUXWqR4M8o0gZYVW12nJwgISk4YgMeun4phx1VygVLGpR1P8XWkt5zZsGYJ1ttoyDFN3Tbjevh3Z3ijlxNXkyIBRYD/May3tSzJNIp28QpLBdOCftvl5kT9h4AJLp5RobZrhebDtsU+NL/M01Qvl5VEeiGUur4uqdGQn6TKyYYZY7zuDz+UXzdJj4i62sUbcY/wDf3HFtVCPMpUlqH+9DOIg4nc7h1+mgEnr+WhS4r2Jt3oCAXaZTIAP/FKXlACbj98MMgqynQE151zHsa077ErWnQCrVeE+OEOP/zEY5rcnjUBoo8avifNu/mGOQ+UC55E56yoa7Dos7KvH6eWxyK9ylu8DybbodhLKNO45oyHICUWjKtm6SwoSLDDsVo5WeFe/RubK5cbmyiM9iCkRae+31xZIyxb2d8pwU0Mivwn4DjYVPGtKyGc5Q2RlotMH9SGEZnjBEjVgGMvVPySlJ5yR7wdIOUfXu76gYfE4kWSm+ky8GlUnKksptfRR80B21Z2nANAd9hS6AXMa8DgZfH7gDYQugunXHnggBR0WEgJu3NQjqhjtRvtE7xva6J5hxyySv0c9ieVIyj9dazyKAhxWN6KDtDxrfIJL3DrZ6e75LaEv0XS9XaBLywyzcmCmvkpIVGERd0pDTmd/YXLuS3xLMuedI6uR4d1AmAPGKkcc7GCdurxYs6lESjq2uRkaoDgMgVavRFfbhntohprN9XF3Ag4OYCmEi5Lnvqx38CDLNpHLPi+lUFCTW5AkPqCP6CqiepPkl97siJU1SHtHXY+7tNDZ+7Khx8HKOOimlJDt9X5Z5r2XOsotaCugW8ItchbLj+m/YtRmXTRlgwkBL3o8p3LaF29pMLMjF6dyg/AlfVHnaHmlN+i3vpAdcvvq3oN0AkkvoyZBNqvknPb/Oq+TOgY5NA3EeacBGeDGL3t+n1xv/q7eadjVZNHKLrnHCA1FJMkiizFL4nwaRVM+05ujxR6YJboPe7D3vwrFxWcKtBGzvOFjotbvxWjAd5+y1G03CL7UGFeNrMhG+sVkxt+wJGitFFTbpi4z+yow+vPj/CF39vZXIMKVo1+UoZaE8mQCtTNMyT3ALkUcqwANaSleQVan/UAPzNkXA/86slnLxmv1UVrGWnoGKspD1JHFUOAoEG0FnrUJVwfXtgeZgKlW0Fsz9lha9+uONBbm7jY9HiXAiAeWydTqD92lo8IwA63NM1havyNkF4nDDrth247V4Vz1j9YMWuxy78z33IUR5ZOKf7YsPZEnfvDvEGw5FUPWQaVywf3uSRAlt122040dn736XKxYhswKEZwccqIf1U+GAxM2z1eHhrYyxezoyTrssBlfjWiULrnU0KTG1Iuu6qZLdxWaQjEqMmgv3+GIvMpd2yZL18dUPnsKabQQ1RAs5415ZQDxyv/0C+IWnaWQiOPJZPoOMMP6IvYtZwjHiBnbcjLM9M23STlntIwRDi5JvoXc/188ZgA=",
        "infoSize": 5634,
        "using": 0,
        "outlineComplete": 1
    }
}
```

Can you post an example response from your bot?
As we can see above, `type` has different values and we need to look into those ones
In addition I think we also need to validate the `outlineVer` as I expected different versions.

edenhaus:
just quickly tested to send different stuff on `getMapInfo_V2` `type` parameter:
- `0` I assume means return all types
- `1` -> will just return this type
- `1,2,5` -> will return all three types and for my bot it's the same data as with type `0`

Looking at your code it looks like you don't have a type `5` but instead `6`

I quickly tested if my bot supports type 6, and it does not.

shmuelzon:
Hey,

> if your bot also supports the getMapInfo command

Doesn't seem that way. I tried to send that command with my `mid` and type `"0"` and got the following error:
```json
{"body": {"code": 20005, "msg": "get map fail"}}
```
So, something is aware of that command and that it's related to a map, but I guess it's not fully implemented or I need to modify the arguments somehow.

> Can you post an example response from your bot?

Sure, here's an example of an `onMapInfo_V2` message (which we, currently, don't handle):
```json
{
    "data": {
        "batid": "icogem",
        "index": "1",
        "info": "XXXXXXX",
        "infoSize": 10682,
        "mid": "1038845207",
        "msgid": "",
        "outlineComplete": 0,
        "outlineVer": "1",
        "serial": "1",
        "type": "0",
        "using": 0
    }
},
```
It's `type` is still `0` here and the `info` payload includes the `1`, `2` and `6` entries. The `outlineVer` is also `1`, like yours which might make sense as it's the same payload format. My `outlineComplete` is `0`, though, not sure what it means...

I cant figure out what that `5` type in your example is. Not a lot of information there...

deebot_client/commands/json/map/__init__.py
```diff
@@ -376,3 +377,24 @@ def _handle_body_data_dict(
...
+
+    @classmethod
+    def _handle_body_data_dict(
+        cls, event_bus: EventBus, data: dict[str, Any]
+    ) -> HandlingResult:
+        event_bus.notify(
```
edenhaus, line 397: We should check here if outline version is one. if not we should return HandlingResult.analyse()

so we prevent strange errors if the version will change in the future.

Also we need to include the type in the MapInfoEvent. I suggest to make a StrEnum of it, where 0 is mapped to All types. This new enum could than be used also above instead the magic 0

tests/commands/json/map/test_init.py
```diff
@@ -488,3 +489,21 @@ async def test_getMapTrace() -> None:
...
     )
+
+
+async def test_getMapInfoV2() -> None:
+    mid = "98100521"
+    info = "KLUv/QRYmQAAW1siMSJdLFsiMiJdLFsiNiJdXbBRuA4="
+    json, firmware_event = get_request_json(
+        get_success_body(
+            {
+                "mid": mid,
```
edenhaus: We should add here the actual response from the bot with all fields (even we are not handling them).
If we add handling for additional fields in the future we make sure that we don't break the current format

src/map/mod.rs
```diff
@@ -229,28 +234,39 @@ impl MapData {
...
+                    .set("x", viewbox.min_x)
+                    .set("y", viewbox.min_y)
+                    .set("width", viewbox.width)
+                    .set("height", viewbox.height)
+                    .set("style", "image-rendering: pixelated")
+                    .set("href", format!("data:image/png;base64,{base64_image}"));
+                document.append(image);
+                viewbox
+            }
+            None => match self.map_info.borrow(py).generate() {
```
edenhaus: As the new map is pure svg we should prioritize it if available.

edenhaus (changes requested):
After this last changes, I will merge it as we can always iterate on and improve it further.
For example in a follow up PR we need to add support for `onMapInfo_V2` otherwise the map will not be updated correctly

edenhaus:
I quickly tested your code with my bot:

This PR:
grafik

Old map with png:
grafik

App:
grafik

(The no mop zone was only created for testing)

Can you please share the same to see how your map looks like in the app

src/map/mod.rs
shmuelzon: Done

tests/commands/json/map/test_init.py
shmuelzon: Done

deebot_client/commands/json/map/__init__.py
shmuelzon: > We should check here if outline version is one. if not we should return HandlingResult.analyse()

Done

> Also we need to include the type in the MapInfoEvent. I suggest to make a StrEnum of it, where 0 is mapped to All types. This new enum could than be used also above instead the magic 0

I'm not sure of this one. The type will either be unhelpful (if it's `"0"`) or, whoever parses the info can already see what's included by looking at the JSON. What's the benefit of adding the types?

shmuelzon:
> I quickly tested your code with my bot:

That's interesting, and also a bit troubling. The map outline (`"1"`) seems to be the same for both our robots, but the `"2"` type is different. In your case it's the area reachable by the robot but, in my case, those are type `"6"`. My type `"2"` is the entire room. I'm guessing so it can be colored similarly as the reachable areas (without the crossed lines) as is shown in the app while your app uses a "generic" background color.

Maybe we can also create a follow-up commit where if type 6 doesn't exist, draw type 2 as if it were type 6. We'll also need to fill in the blanks (create a filled polygon) throughout the entire floor plan. We can do the last part regardless since if the information will be available, it'll just be overlayed on top with the rooms so it won't be visible.

This PR:
image

From the app (technically, I have it rotated in the app, but we removed support for that for now):
image


BTW - The above also shows something that I was wondering if it's a bug or just behaves differently between the different robots. You'll notice that the app doesn't show traces for when the robot it just positioning it self, e.g., moving to from the station, it only shows traces where it's actually cleaning. The SVG map, however, does show those too.

shmuelzon:
>For example in a follow up PR we need to add support for onMapInfo_V2 otherwise the map will not be updated correctly

IIRC, when I tried the map selection thing on my robot, I got multiple `onMapInfo_V2`. The problem was that I didn't know if it was for the active map or not since we get the map ID for the `GetCachedMapInfo` command and use that to get the map info for the map ID that was returned there.

edenhaus (approved):
Thanks @shmuelzon 👍 

The other things we can handle in follow up PRs :)

Looking about the map from the apps I would say type `2` is "rooms". In your case the whole room will be shown as room and in my case only the reachable part of it. But I think thats the correct way forward as on my map only the reachable part is colored

shmuelzon:
> Looking about the map from the apps I would say type 2 is "rooms". In your case the whole room will be shown as room and in my case only the reachable part of it. But I think thats the correct way forward as on my map only the reachable part is colored

That's just it, in terms of how it's rendered, your "2" and my "6" are displayed as solid infill polygons with the "room" color. That's why I think they should be treated the same. What's displayed as a crossed-line infill on my app is the "extra data" that's nice to have. I'll play around with it soon and see how it turns out.

BTW - any objections to adding color to the SVG-based maps? They'll differ from the existing ones but will look nicer and more similar to the app

edenhaus:
> BTW - any objections to adding color to the SVG-based maps? They'll differ from the existing ones but will look nicer and more similar to the app

No we can add colors in follow up PRs.
I just want to make sure it looks good for all version of the map. For example I'm fine to change the wall color to gray as it's in your app but than we should do it also in the png to have a consistent map experience.

Sanji78:
Well done @edenhaus  @shmuelzon. In which version of HA will this be available?

robin-thoni:
Thanks @shmuelzon for finishing what I started :)

Thanks @edenhaus for the numerous reviews, suggestions and overall work!

@Sanji78 We first need a release of this library, then a bump in HA, like [this one](https://github.com/home-assistant/core/pull/154154) (I don't think this requires any modification in the way HA consumes the library, right?). If @edenhaus can release the new version this week, I guess we should be able to have it in `2025.11.0` (so roughly, two weeks from now)?

Sanji78:
Thanks for clarification. @shmuelzon  @edenhaus  did you also add support for rooms as attributes of the vacuum entity?

shmuelzon:
> Thanks for clarification. @shmuelzon @edenhaus did you also add support for rooms as attributes of the vacuum entity?

That was already added as a part of https://github.com/DeebotUniverse/client.py/pull/1186

Sanji78:
@shmuelzon  well done!!!

edenhaus:
Please don't use a merged PR to discuss things unrelated to it.
Also with all changes they will be available in HA when the version is bumped and I release a new version, when I think the version and it's feature are stable enough

robin-thoni:
@edenhaus @shmuelzon I just tried the `16.1.0` release, which, as I understand, should contain that fix, but it still doesn't work in HA:
image
image

I'm using the `ecovacs` integration from `2025.10.4` w/ the `deebot-client` requirement bumped to `16.1.0`, without any other modification. Does it actually require patches in the HA integration?

edenhaus:
@robin-thoni Don't comment on a merged PR if you were not part of it, and don't ping members asking for help.
Please don't blindly install a different version, and depending on your model, the new map needs to be activated.

PR #1182, opened 2025-09-24
Handle new MapSetV2 data format

shmuelzon:
The data format for `MapSetV2` from my robot (X9 Omni) seems a bit different from what's currently supported.
For starters, it's `zstd` compressed and not `lzma`. Second, the payload includes an extra field:
```json
[
    [
        "0",
        "1", #<- This
        "-4814",
        "12059",
        "-4814",
        "7768",
        "-3948",
        "7768",
        "-3948",
        "12059"
    ],
    [
        "1",
        "1", #<- And this
        "3315",
        "3754",
        "3353",
        "-655"
    ]
]
```
While the existing messages had an ID as the first field and then X, Y pairs for points, my robot always included a second field after the ID, which always seems to be `"1"` for me, before the X, Y pairs.

For now, I opted to simply check if the number of fields, following the ID is even or not (as we expect pairs) and, if not, just remove the extra field. Seemed simpler, and more robust, as opposed to checking which compression algorithm was used or marking it in the capabilities or something. I saw this "extra" field for both virtual walls and no-mop zones so all are handled and placed correctly on the map now (once the map implementation will be ready).


tests/commands/json/test_map.py
```diff
@@ -532,6 +532,47 @@ async def test_getMapSetV2_virtual_walls() -> None:
     )
 
 
+async def test_getMapSetV2_virtual_walls_new_zstd_with_extra_entry() -> None:
```
edenhaus: Please migrate the test above to a parameterized one and add this one there

edenhaus (changes requested):

tests/commands/json/test_map.py
shmuelzon: Done

edenhaus (approved):
Thanks @shmuelzon 👍

PR #1186, opened 2025-09-26
Add rooms support for robots without GetMapSubSet

shmuelzon:
Newer robots (at least X8 and X9) don't support the `GetMapSubSet` command and so the flow for getting the rooms subset ID with `GetMapSetV2` and calling `GetMapSubSet` to get the names and coordinates of each room doesn't work.

The `GetMapSetV2` command response already includes the names of each room and the `MapInfoV2`, which isn't merged/supported yet, includes the coordinates. Since I didn't see where the coordinates are actually used, neither here nor in the HA integration, I figured adding support for their names would be helpful and also remove the `Handler for message getMapSubSet: {'code': 20003, 'msg': 'rcp not support'} returned no response. This is a bug should not happen. Please report it.` error message.

Once `MapInfoV2` is merged, I think we'll need to refactor a few things to include the coordinates information for the rooms since unlike the existing robots, we'll need to parse and combine two different messages to get that information.


deebot_client/commands/json/map.py
```diff
@@ -317,12 +329,63 @@ def _get_subset_ids(
...
+        event_bus: EventBus,
+        data: dict[str, Any],
+        map_type: MapSetType,
+        subsets: list[list[str]],
+    ) -> HandlingResult:
+        subset_ids = [int(subset[0]) for subset in subsets]
+        result = cls._get_handling_success_with_subset_command_args(data, subset_ids)
+
+        # there are two versions of this message, depending on the number of values
+        if len(subsets[0]) == 10:
```
shmuelzon: What if the subsets is an empty list, this will raise an exception.

tests/commands/json/test_map.py
```diff
@@ -485,6 +488,52 @@ async def test_getMapSetV2_rooms() -> None:
...
+                "index": 1,
+                "subsets": subsets_comp,
+                "infoSize": 780,
+            }
+        )
+    )
+    events: list[Event] = [firmware_event]
+    for subset, room_name in zip(subsets, rooms_names, strict=False):
+        events.append(MapSubsetEvent(subset, set_type, "", room_name))
+    events.append(MapSetEvent(MapSetType(set_type), subsets))
```
shmuelzon: Your change has reversed the order between the `MapSetEvent` event and the `MapSubsetEvent` events. Because of this, the `RoomsEvent` message is never sent due to the logic at `deebot_client/map.py` where the `RoomsEvent` message is published only if we got `MapSubsetEvent` to match the number of rooms, but we don't know the number until `MapSetEvent` is sent.

edenhaus: Thanks for letting me know. I did another round of refactoring. Can you please verify if I didn't miss anything :)

shmuelzon: Hey, I can confirm that now I do get the `RoomsEvent`, so thanks for that!

I see that it's now sent directly from the `MapSetV2` and not from the "generic" `Map` class (or now `MapRoomHandling`). If/When we want to also include the coordinates (not sure how critical that really is), we might want to move that logic to `MapRoomHandling` to aggregate the information and, once it's all available, publish the `RoomsEvent`.

edenhaus: We can add support for coordinates later in a follow-up PR and refactor the code there. I want to avoid the need for events to be in a specific order to work (if possible) as it makes the code only more complex.

shmuelzon: Totally agree, thanks!

edenhaus (approved):
Thanks @shmuelzon 👍

shmuelzon:
> Thanks @shmuelzon 👍

I think that, at this stage, you've put in more work in this PR than I did, so thank you 😉

Sanji78:
@shmuelzon 2 of my rooms have blank name (maybe there's a missing Italian translation) ? If you tell me where is this handled, I can push a PR for fixing it. Thanks

shmuelzon:
> @shmuelzon 2 of my rooms have blank name (maybe there's a missing Italian translation) ? If you tell me where is this handled, I can push a PR for fixing it. Thanks

Hey, the event (that HA listens on) is sent from https://github.com/DeebotUniverse/client.py/blob/ec2c2ba4993dd05a275914d304aa91ba7cf9d626/deebot_client/commands/json/map/__init__.py#L289

Sanji78:
@shmuelzon  thanks. Do you know where names are stored? In this client.py project I don't find any room names... is it coming from the MQTT packet of the robot or it's part of HA ecovacs component?

Sanji78:
Ok... I found out that names is coming directly from MQTT message. For some reason the standard Bathroom room name (when you click on bathroom icon in the map editor of the Ecovacs map) is seen as blank... but you can edit the label on Ecovacs map editor.. and when you do it, the name is coming...

Thank you

PR #1204, opened 2025-10-04
feat: add support for GpsPositionEvent

JeffResc:
The Ecovacs Goat A3000 ([cr0e4u](https://github.com/DeebotUniverse/client.py/blob/dev/deebot_client/hardware/cr0e4u.py)) has an `onGpsPos` event that emits the GPS coordinates of the lawn mower. This PR adds support for that event so that you may subscribe to and use that information


edenhaus (approved):
Thanks @JeffResc 👍

PR #1221, opened 2025-10-11
Improve Deebot Mini support (55uoqe)

aronnebrivio:
Improve Debot Mini support by adding a dedicated configuration.

# Vacuum features
- add Border Spin feature and capability

# Station features
- add "drying mop" handling

# General
- new error codes and messages for clean and dirty water tanks errors

# 55uoqe improvements
- round mop and dust bag lifespan capabilities
- work modes clean capabilities 
- border spin capability
- station capabilities 

Screenshot 2025-10-11 193355
Screenshot 2025-10-11 193405
Screenshot 2025-10-11 193419

Note: I have some stashed work on HA Ecovacs integration to be submitted once the client is updated

PR #1224, opened 2025-10-11
Add border spin command

aronnebrivio:
Border spin is the way bots like the Mini, without an extendable mop, can reach the edges by tilting itself

(split from https://github.com/DeebotUniverse/client.py/pull/1221)


edenhaus (commented):
Thanks @aronnebrivio 👍

edenhaus (approved):

PR #1227, opened 2025-10-11
Use path instead polygons

edenhaus:
New map svgs are more than **20%** smaller
@shmuelzon Path should be a drop in replacement for polygons but in case something is not working please let me know


shmuelzon:
> @shmuelzon Path should be a drop in replacement for polygons but in case something is not working please let me know

Tested this branch with my robot and the new map looks, visually, the same. Nice!

PR #1231, opened 2025-10-12
Fix map for bots, which don't support BlockLine

edenhaus:


src/map/map_info.rs
```diff
@@ -115,7 +109,39 @@ impl MapInfo {
...
+                    vec![CSSClass::FillNone, CSSClass::OutlineStroke],
+                    false,
+                ),
+            ]
+        } else {
+            [
+                (MapInfoType::Outline, vec![CSSClass::RoomUnreachable], true),
+                (
+                    MapInfoType::Room,
+                    vec![CSSClass::RoomReachable, CSSClass::OutlineStroke],
```
shmuelzon, line 135: Hey, it's probably a matter of personal taste but I think the stroke here is redundant and should be removed to look more similar to the maps generated by the new/full MapInfo_V2

Image
Compared to:
Image

```suggestion
                    vec![CSSClass::RoomReachable],
```

edenhaus: I added it on purpose to meet the new app look, as you can see in https://github.com/DeebotUniverse/client.py/pull/1177#issuecomment-3393500863
The app adds a border to it (even that I just saw that is a little bit smaller)
So that you know, Ecovacs is not drawing the walls correctly. For example, the kitchen (left down corner) is separated by a wall with a door from the hallway. Also the hallway is straight and will not increase/decrease in the middle.
Also the winter garden (right) is separated by a wall with a door

shmuelzon: > The app adds a border to it (even that I just saw that is a little bit smaller)

Right, I see that now, it was just different than what I'm used to from my app.

> So that you know, Ecovacs is not drawing the walls correctly. 

Yeah, I guess it all boils down to the algorithm they use to map the floor layout. Mine's generally fine but the robot doesn't totally distinguish between furniture that's right by the wall and the wall itself so some places are a bit wonky.

Thanks!

PR #1233, opened 2025-10-12
Add support for onMapInfo_V2 message

shmuelzon:
(Cannibalizing the same PR...)

I've noticed that when changing the map via the app (might also be from the library, haven't tried), the SVG map isn't updated to reflect the selected map. To support that, I've added a handler for the onMapInfo_V2 message to get the new map information that is later used to draw the map.

I was reluctant to add it before because sometimes I get this message with an irrelevant map ID (not one of the two maps I've created). However, after getting those irrelevant messages, the final one that's received is of the selected map so everything's fine at the end. Another thing that I did notice today when looking at it, is that the `outlineVer` of these irrelevant messages is `0` where we "officially" only support `1` so they will be filtered away by the handled. Not sure if this is good or bad since it also results with an log message that the messages couldn't be handled. Let me know if that's fine or you want me to silently ignore that message.


edenhaus:
I'm against this PR (at least in this form), as we shouldn't usually request commands from messages. I just added it in case the "on" message lacks relevant information. At this point, it is only implemented for `onMapSet_V2`.
Also, I just checked my bot, and I get `onMapInfo_V2` and `onMapSet_V2` when I change something that affects these messages (like changing the map). So we can't move it and request a manual refresh for all bots.
It would be strange if Ecovacs changed, so these messages would not be sent anymore.
Please analyze it in more detail... Could you compare it with the message in the tests? Maybe the format changed, and we can detect when we manually need to call the refresh commands.
Is the app always manually calling the commands?

FYI: When manually requesting the command `getMapSet_V2` (for all types), it will also request the Mapsubset commands, which I only want to request when really needed and until now no bot needs these manual requests.

shmuelzon:
Hey, I updated/modified the pull request to rely on the `onMapInfo_V2` message which I agree is much more cleaner than what I suggested before. I described above why I was hesitant about using it but I guess it's OK now.

As for the issue of virtual walls not appearing after selecting a different map, let me know if you want to handle it here or if I should open a separate bug/PR. The gist of it is this:
* I have only a single virtual wall in my map
  * My robot doesn't have any other `MapSetType`s as the rooms are handled differently and I don't have any no-mop zones
* When starting my test script, everything's enumerated and displayed correctly in the map
* I then switch to my other (mostly empty, test) map which doesn't have any virtual walls
  * This clears the existing virtual wall entry subset as get subsets returned an empty list
* I then switch back to the main map that, in turn, get's all the map set information
* The single virtual wall I have is also received and an event is to be published on it
  * The event is filtered out since it's the same event that was sent when the original map was loaded
  * This means that the new map is drawn without the virtual wall

If I had a virtual wall in my second map, then everything would be fine since the event would be different each time and not filtered. Likewise, if I had a second virtual wall in my map, it would also work since each event would be be different when sending the first subset, then the second, (map change) then the first again.
Not sure how you want to handle this, my feeling is that `MapSubsetEvent` shouldn't be filtered but there's no mechanism for that so we can either add some `force` flag when publishing the message or add some unique field to the event so it'll be different each time. Or something completely different that I'm not thinking of :)

Let me know what you think...

deebot_client/messages/json/map/__init__.py
```diff
@@ -65,3 +66,24 @@ def _handle_body_data_dict(
...
+    @classmethod
+    def _handle_body_data_dict(
+        cls, event_bus: EventBus, data: dict[str, Any]
+    ) -> HandlingResult:
+        """Handle message->body->data and notify the correct event subscribers.
+
+        :return: A message response
+        """
+        if data.get("outlineVer") != "1":
+            return HandlingResult.analyse()
```
edenhaus:
```suggestion
        if (outline_version := data.get("outlineVer"))  == "0":
            # Skip it as it will be sent for not active maps
            return HandlingResult.success()
        elif outline_version != "1":
            # Unsupported veriosn
            return HandlingResult.analyse()
```

deebot_client/messages/json/map/__init__.py
```diff
@@ -65,3 +66,24 @@ def _handle_body_data_dict(
             HandlingState.SUCCESS,
             {"map_id": map_id, "values": values},
         )
+
+
+class OnMapInfoV2(MessageBodyDataDict):
```
edenhaus, line 71: Please let `GetMapInfoV2` inherit from this class to avoid duplicate code

edenhaus (changes requested):
For the issue about duplicate events, please create a separate issue for it as it's unrelated to this PR and I need to think about how to fix it.
Maybe have internal and external events, where internal events are not filtered

deebot_client/messages/json/map/__init__.py
shmuelzon: Done (slightly modified)

deebot_client/messages/json/map/__init__.py
shmuelzon: Done

edenhaus (approved):
Thanks @shmuelzon 👍

PR #1242, opened 2025-10-13
Upgrade to Rust 2024 edition

edenhaus:


edenhaus:
Upgrade with the next breaking change version

PR #1252, opened 2025-10-18
Add support for map rotation

shmuelzon:
This PR adds support for rotating the map according to the user's preference in the app. The value is a part of the cached map info structure so, if it's there, we include it in the map information, otherwise, we use the default 0 degrees.

This time I chose not to make use of the built-in SVG transformations as we still needed to calculate the new viewbox, taking into consideration the rotation, so we needed to calculate the new points anyway. This made the code more complicated as everything was done twice. In addition, I opted to support only 90-degree turns to avoid "heavy" trigonometry functions, I'll soon find out how much it hurt performance when the benchmark pipeline runs.

A few notes to consider when reviewing:
* I'm sure there are better ways to "Rustify" my code, any suggestions are welcome
* There are 3 places that need to handle the rotation where each does it a bit differently (different axes, offsets, etc.) so it might seem a bit cumbersome. I tried to separate those to functions to make the code a bit easier to read
* The map info version of the map calculated all of the points when the information has arrived but the rotation angle might be set before/after. To handle that I save the "raw" X,Y values from the robot and only calculate the new positions when generating the SVG. This is a bit disappointing as it's now calculated each time the SVG is generated even though it's not likely to change, but I don't think I can rely on the order of the messages. It's also, for better or worse, similar to how the background image is generated
* Regarding the tests, I currently generate each map data with all rotations (using the private `_map_data.set_rotation_deg()`), this seemed a bit wrong to me but I though it would cover more than adding one type of rotation to each map type (map_info and background image) or adding 8 more map data items to cover all the possibilities. Let me know what you think of it

As always, thank you for your continued work on this, I am now hooked on seeing the map updated in HA when the robot is cleaning :)

Edit - Since the map generation tests were changed to support rotation, the benchmark doesn't verify them against `dev` so I manually compared with https://codspeed.io/DeebotUniverse/client.py/runs/68eea224abcd9f647da55998:
| test | dev | PR (with 0 degree rotation) |
| - | - | - |
| test1 | 26.3 ms | 26.7 ms |
| test2 | 22.1 ms | 22.4 ms |
| test3 | 8.5 ms | 8.5 ms |
| test4 | 7.2 ms | 7.2 ms |

So it seems there is some loss on maps that are based on the background image. I'll try to inline the functions as you've done elsewhere and see if that makes any difference.

Edit 2 - It doesn't... I tried changing the functions to closures that would avoid the `match` on each iteration but the end result is still the same. Weird...


tests/test_map.py
```diff
@@ -206,22 +206,28 @@ async def test_empty_maptrace(
...
-        id=f"{filename}-{module.DEVICE_CLASS}",
-    )
+    return [
+        pytest.param(
+            module.DEVICE_CLASS,
+            module.EVENTS,
+            angle,
+            id=f"{filename}-{module.DEVICE_CLASS}-{angle}",
+        )
+        for angle in [0, 90, 180, 270]
```
edenhaus: We should only create rotation tests for devices that support it and not everyone.
Especially, these tests should be maps that are also generated from the actual model.

edenhaus (commented):
I was not reviewing it in detail, but merging it as it is will increase the map generation for old devices up to 30%. Can you maybe ask AI to improve the code?
Just so you know, the performance tests are executed on cloud servers, which are much more powerful than the end devices users will use. I moved the map code to Rust, as some users had issues with the Python code back in the day.

tests/test_map.py
shmuelzon: Hey, it's closer to 3%, even less, not 30%.
If we assume that older models don't support rotation anyway, then I can remove the support for it when generating the background image but I don't know if there are any versions in between that use the background image but also allow rotation in the app, and those will break, I can remove and wait to see if anyone files a bug 🤷 

BTW - It may also be, as you've found from your robot, that other robots can switch their hardware file to use the new map API and cut the SVG generation time to a third.

Please let me know your thoughts and I'll modify accordingly.

Thanks!

deebot_client/events/map.py
```diff
@@ -113,6 +113,7 @@ class Map:
     name: str
     using: bool
     built: bool
+    angle: int = field(default=0, kw_only=True)
```
edenhaus:
```suggestion
    angle: int
```

tests/test_map.py
edenhaus: I have considered the worst case:
- yna5xi: 33.3/26.3 => ~26%
- kr0277: 29/22.1 => ~37%

Only bots that have been recently updated support the new format. So any older bot will never get an update.
Are we sure that the angle can only be 0, 90, 180, 270? If not, we should create a warning log entry, so users will report it if there are different values.

shmuelzon: My app only allows 90 degree rotation
IMG_7530

I can add an error/warning just the same.

What do you say about rotation of the background image? Should I remove that as well with an error if it's ever attempted?

deebot_client/messages/json/map/cached_map_info.py
```diff
@@ -40,6 +40,7 @@ def _handle_body_data_dict(
                 name=map_info.get("name", ""),
                 using=map_info["using"] == 1,
                 built=map_info["built"] == 1,
+                angle=map_info.get("angle", 0),
```
edenhaus: Check the angle here and create a warning if it's not a valid one. In case of an invalid one, set it to 0

src/map/background_image.rs
```diff
@@ -29,6 +29,58 @@ const MAP_IMAGE_PALETTE_TRANSPARENCY: &[u8] = &[0u8, 255, 255, 255, 255, 255];
 const MAP_PIECE_SIZE: u16 = 100;
 pub(super) const MAP_MAX_SIZE: u16 = 8 * MAP_PIECE_SIZE;
 
+type PiecePositionCalculator = Box<dyn Fn(usize) -> (u16, u16)>;
```
edenhaus: Remove rotation support for now

src/map/map_info.rs
```diff
@@ -82,7 +82,7 @@ impl MapInfo {
         }
     }
 
-    pub(super) fn generate(&self) -> MapInfoGenerateResult {
+    pub(super) fn generate(&self, rotation_deg: i16) -> MapInfoGenerateResult {
```
edenhaus: Create an enum for the rotation with the 4 valid cases and use it directly in Python. See `PositionType` as an example

tests/test_map.py
edenhaus: I would remove the background image rotation for now (as I'm expecting that it will not be used). Also, my bot is not even showing the rotation setting in the app.

As we are using an enum, we should create a log entry for invalid angles and set it to 0.
Please also remove the rotation from all tests expect your bot as the other don't support rotation

tests/test_map.py
```diff
@@ -242,6 +249,8 @@ async def test_fn() -> str | None:
             event_bus.notify(event)
 
         await block_till_done(event_bus)
+
+        map_obj._map_data.set_rotation_deg(angle)
```
edenhaus: We should not set the rotation here. Instead set the angle in the correct event

edenhaus (changes requested):

tests/test_map.py
shmuelzon: Done (I hope the enum is what you were thinking of)

deebot_client/messages/json/map/cached_map_info.py
shmuelzon: The check/log and default values are implemented in Rust

src/map/background_image.rs
shmuelzon: Done

src/map/map_info.rs
shmuelzon: Done

tests/test_map.py
shmuelzon: Done

tests/messages/json/map/test_cached_map_info.py
```diff
@@ -38,6 +39,7 @@
                     "built": 1,
                     "name": "Erdgeschoss",
                     "isFastBuilding": 1,
+                    "angle": 90,
```
edenhaus:
```suggestion
```

Don't modify this test as I want to check the data as much possible to the original.
Please add a new tests with the result of your bot

src/map/mod.rs
```diff
@@ -116,6 +122,46 @@ impl PositionType {
...
+            }
+        }
+    }
+}
+
+#[pymethods]
+impl RotationAngle {
+    #[staticmethod]
+    fn from_int(value: i16) -> Self {
+        RotationAngle::from(value)
```
edenhaus: Please move the mapping from above here and raise instead of creating a log entry.

The reason behind this is that I had a discussion today with a colleague about it and it's easier to add new functionality instead of providing the wrong one (Using 0 instead the correct angle)

I don't expect that different values currently exist but we will see if users are reporting errors

edenhaus (changes requested):
It looks amazing some small improvement and we can merge :)
I will bump afterwards the lib and try to include in HA 2025.11

src/map/mod.rs
shmuelzon: Sure, done

tests/messages/json/map/test_cached_map_info.py
shmuelzon: done

edenhaus (approved):
Thanks @shmuelzon :+1:

PR #1273, opened 2025-10-28
Prevent old pyo3-log version

edenhaus:


edenhaus:
codspeed slowdown is for sure unrelated

PR #1279, opened 2025-10-31
Add Autoempty2

ha-newbie:
When selecting from “select” in HA, it calls executeCommand with the selected string as an option. However, auto empty requires two parameters, causing an error in SetAutoEmpty. (GetAutoEmpty works without any issues.)

```
  File "/usr/local/lib/python3.13/site-packages/deebot_client/commands/json/auto_empty.py", line 33, in __init__
    params["enable"] = int(enable)
                       ~~~^^^^^^^^
ValueError: invalid literal for int() with base 10: 'smart'
```

By the way, the Deebot T50 OMNI sets enable=0 when frequency=manual.
Therefore, I added SetAutoEmpty2, which limits the received parameter to frequency only. It sets enable=0 only when frequency=manual and enable=1 otherwise.
Tested in HAOS (Deebot Client 16.0.0, HA Core 2025.10.4) with translation added.

If you see any strange code or know a better way, please leave a comment.

### DIFF
--- a/deebot_client/commands/json/auto_empty.py
+++ b/deebot_client/commands/json/auto_empty.py
@@ -29,7 +29,8 @@ class SetAutoEmpty(ExecuteCommand):
             frequency = get_enum(Frequency, frequency)
 
         if isinstance(enable, str):
-            if frequency is None: frequency = get_enum(Frequency, enable)
+            if frequency is None:
+                frequency = get_enum(Frequency, enable)
             enable = frequency != "manual"
 
         params: dict[str, Any] = {}
--- a/deebot_client/commands/json/auto_empty.py
+++ b/deebot_client/commands/json/auto_empty.py
@@ -28,28 +28,16 @@ class SetAutoEmpty(ExecuteCommand):
         if frequency is not None and not isinstance(frequency, Frequency):
             frequency = get_enum(Frequency, frequency)
 
+        if enable is not bool and frequency is None:
+            frequency = get_enum(Frequency, enable)
+            if frequency == “manual”:
+                enable = 0
+            else:
+                enable = 1
+
         params: dict[str, Any] = {}
         if enable is not None:
             params["enable"] = int(enable)
-        if frequency:
-            params["frequency"] = frequency.value
-        super().__init__(params)
-
-
-class SetAutoEmpty2(ExecuteCommand):
-    """Set auto empty command without enable."""
-
-    NAME = "setAutoEmpty"
-
-    def __init__(self, frequency: Frequency | str | None = None) -> None:
-        if frequency is not None and not isinstance(frequency, Frequency):
-            frequency = get_enum(Frequency, frequency)
-
-        params: dict[str, Any] = {}
-        if frequency == "manual":
-            params["enable"] = 0
-        else:
-            params["enable"] = 1
         if frequency:
             params["frequency"] = frequency.value
         super().__init__(params)
--- a/deebot_client/commands/json/auto_empty.py
+++ b/deebot_client/commands/json/auto_empty.py
@@ -30,10 +30,7 @@ class SetAutoEmpty(ExecuteCommand):
 
         if enable is not bool and frequency is None:
             frequency = get_enum(Frequency, enable)
-            if frequency == "manual":
-                enable = 0
-            else:
-                enable = 1
+            enable = 0 if frequency == "manual" else 1
 
         params: dict[str, Any] = {}
         if enable is not None:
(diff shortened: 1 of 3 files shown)


deebot_client/commands/json/auto_empty.py
```diff
@@ -34,3 +34,22 @@ def __init__(
         if frequency:
             params["frequency"] = frequency.value
         super().__init__(params)
+
+
+class SetAutoEmpty2(ExecuteCommand):
```
edenhaus: Don't create a new command, when using the same one. Extend the already existing one

edenhaus (changes requested):
Please also add tests for it

deebot_client/commands/json/auto_empty.py
ha-newbie: I added `str` to the expected types for `enable` and modified it to handle exceptions when `str` is passed to `enable`.
Is this acceptable?
(Please ignore the large number of garbage commits...)

ha-newbie:
I've also added tests. Please let me know if there are any mistakes or omissions.

deebot_client/commands/json/auto_empty.py
```diff
@@ -23,11 +23,16 @@ class SetAutoEmpty(ExecuteCommand):
     NAME = "setAutoEmpty"
 
     def __init__(
-        self, enable: bool | None = None, frequency: Frequency | str | None = None
+        self, enable: bool | str | None = None, frequency: Frequency | str | None = None
```
edenhaus, line 26: When is `enable` a `str`?

ha-newbie: When selecting a dropdown entity in Home Assistant (though what actually selected is `frequency`)

deebot_client/commands/json/auto_empty.py
```diff
@@ -23,11 +23,16 @@ class SetAutoEmpty(ExecuteCommand):
...
     ) -> None:
         if frequency is not None and not isinstance(frequency, Frequency):
             frequency = get_enum(Frequency, frequency)
 
+        if isinstance(enable, str):
+            if frequency is None:
+                frequency = get_enum(Frequency, enable)
```
edenhaus, line 33: This is wrong... You are passing in the frequency as enable param...

deebot_client/commands/json/auto_empty.py
edenhaus: Please pass them in as frequency param and not the enable one also we should not override the enable param if set. Only set it when it is None

edenhaus (changes requested):

ha-newbie:
The fundamental problem is that the `str` option selected via the select entity in Home Assistant is passed to executeCommand without argument specification.
To resolve this:
1.  Enable argument specification in the integration, or swap frequency as the first argument (but it must automatically assign even when enable is None -> disrupt existing behavior).
2.  Enable passing multiple arguments from a single option in the integration (should be the best approach).
3. Create a different command class altogether (rejected)
4. Write logic to treat the `str` passed to `enable` as `frequency` if forced (doesn't disrupt expected behavior but isn't ideal programmatically)

I didn't know how to implement methods 1 or 2, so I proposed 3 or 4. However, I'll reconsider 2. Any assistance or better solutions would be greatly appreciated.

deebot_client/commands/json/auto_empty.py
edenhaus:
```suggestion
        self, enable: bool | None = None, frequency: Frequency | str | None = None
```

deebot_client/commands/json/auto_empty.py
```diff
@@ -23,11 +23,16 @@ class SetAutoEmpty(ExecuteCommand):
...
+            enable = frequency != "manual"
```
edenhaus, line 34:
```suggestion
        if enable is None:
            enable = frequency is not Frequency.MANUAL:
```

edenhaus (changes requested):
> The fundamental problem is that the str option selected via the select entity in Home Assistant is passed to executeCommand without argument specification.

That's not correct. The capabilities for station auto_empty are defining the arguments `[bool | None, auto_empty.Frequency | str | None]` so both need to be provided. In HA, we still need to add support for it, but there you can just pass `None, frequency`

ha-newbie:
What I meant was the behavior within the ‘select’ integration. This is not about this repository.
Naturally, commands like `sendcommand` work fine, but the options selectable within the integration are limited to a single `str`. With the current code, specifying arguments is not possible.
I believe I need a logic that takes `key` and `option` as input and returns the parameters of `dict`.
(My model needs both `enable` and `frequency`.)
https://github.com/home-assistant/core/blob/4dfb6e4983686c51f57398b00be4b0fa98bb432d/homeassistant/components/ecovacs/select.py#L114-L116
```
    async def async_select_option(self, option: str) -> None:
        """Change the selected option."""
        await self._device.execute_command(self._capability.set(option))
```

However, I understand that changing this repo is not the correct way, so I will close this PR for now. (I will create a separate PR for the `"manual"` addition.)
If I can't create the PR in HA myself, I'll raise a feature issue, so I'd appreciate any help.

PR #1281, opened 2025-11-01
Add support for colorized rooms

shmuelzon:
This PR adds support for coloring rooms/areas in the SVG map and it's only available for robots that support MapInfo.

Most of the elements remain the same as before. The main change is with the `Room` items which now need to be drawn twice. Once with the color of the room and another time with the striped pattern overlayed on top of it. I couldn't find a way to have both a background color and an overlay on the same path. The new styles and striped line definitions are only added when needed to keep the SVG small.

Examples:
| Before | After |
| - | - |
| image | image |
| image | image |

Note that the background image-based maps have changed as a result but their content is the same and only the `defs` element was moved to simplify the code.


src/map/map_info.rs
```diff
@@ -8,9 +8,21 @@ use pyo3::prelude::*;
...
+    ViewBox,
+    OrderSet<CSSClass>,
+    Vec<Box<dyn svg::node::Node>>,
+)>;
+
+const ROOM_COLORS: [CSSClass; 4] = [
+    CSSClass::RoomColor1,
+    CSSClass::RoomColor2,
+    CSSClass::RoomColor3,
+    CSSClass::RoomColor4,
```
edenhaus: Why only 4 rooms? I'm expecting a normal flat has often more rooms

Can we have a different blue? It's nearly the same as the wall color and the other colors look a lot better. Maybe even don't have a blue et all

src/map/map_info.rs
```diff
@@ -115,51 +127,92 @@ impl MapInfo {
...
+                .add(
+                    Line::new()
+                        .set("x1", 0)
+                        .set("y1", 0)
+                        .set("x2", 0)
+                        .set("y2", 2)
+                        .set("stroke", "rgba(0, 0, 0, 0.2)")
+                        .set("stroke-width", 1),
+                );
+            added_defs.push(Box::new(pattern));
```
edenhaus: This should go into styles.rs as it's related to them.

src/map/map_info.rs
```diff
@@ -115,51 +127,92 @@ impl MapInfo {
...
+        }
+
+        Some((svg_elements, viewbox?, used_styles, added_defs))
     }
 
-    fn get_order(&self) -> [(MapInfoType, Vec<CSSClass>, bool); 3] {
+    fn get_order(&self) -> Vec<(MapInfoType, Vec<CSSClass>, bool, bool)> {
```
edenhaus: We should create a struct (instead of a tuple) for better readability

src/map/map_info.rs
shmuelzon: I took the 4 colors that seem to repeat themselves from the app. The app also has the walls as gray (see https://github.com/DeebotUniverse/client.py/pull/1177#issuecomment-3393566337) so that's another option but then it'll be different than what your app shows.

If we're good with deviating from the app, than I'd suggest adding more colors to reduce the risk of two rooms that have a shared edge having the same color. I couldn't understand yet the logic of the app to decide the color for each room but somehow (might be by chance) that doesn't happen even with "only" 4 colors.

Anyway, let me know what you prefer and I'll do that

src/map/map_info.rs
shmuelzon: Done something, tried to keep it similar to the styles and correlate the dependencies between the two so it will be simpler to add more later, if needed. Hope it's to your liking

src/map/map_info.rs
shmuelzon: Done

src/map/map_info.rs
edenhaus: I don't remember which command returns the connection between rooms, but maybe the app uses it to pick the correct color. Nevertheless, we should add more colors and make it better :)
We can try to go with the gray walls for all and see how it looks :)

Not sure if you want to add also the room colors for the png based map. If yes, you should only need to modify https://github.com/DeebotUniverse/client.py/blob/dev/src/map/background_image.rs#L70-L74

src/map/style.rs
```diff
@@ -1,16 +1,65 @@
...
+
+type DefinitionFactory = fn() -> Box<dyn svg::node::Node>;
+
+fn get_definitions() -> &'static HashMap<Definition, DefinitionFactory> {
+    static DEFINITIONS: OnceLock<HashMap<Definition, DefinitionFactory>> = OnceLock::new();
+    DEFINITIONS.get_or_init(|| {
+        HashMap::from([(
+            Definition::DiagonalStripes,
+            (|| {
+                Box::new(
```
edenhaus: Why are you wrapping everything in a Box? I removed some of them already in your previous PRs

src/map/style.rs
```diff
@@ -80,6 +161,15 @@ pub(super) struct CSSEntry {
     pub identifier: &'static str,
     pub value: &'static str,
     pub class_name: &'static str,
+    pub required_defs: &'static [Definition],
```
edenhaus: We don't need to have a list if we only have at most one element

src/map/style.rs
```diff
@@ -101,4 +191,11 @@ mod tests {
             );
         }
     }
+
+    #[test]
+    fn test_get_definitions_has_all_members() {
+        for variant in Definition::iter() {
+            let _node = get_definition(&variant);
```
edenhaus: Please also check if the ids are unique

src/map/map_info.rs
shmuelzon: I switched the walls to gray but, if you don't like it, `#c6d5eb` seems like a good replacement from the RoomColor1's blue. Also added two more colors:
| `ilt3k8` | `kr0277` |
| - | - |
| image | image |

> Not sure if you want to add also the room colors for the png based map

I tried adding the following change:
```patch
diff --git a/src/map/background_image.rs b/src/map/background_image.rs
index 99a5101..7f29d6d 100644
--- a/src/map/background_image.rs
+++ b/src/map/background_image.rs
@@ -20,8 +20,14 @@ const MAP_IMAGE_PALETTE: &[u8] = &[
     0x1a, 0x81, 0xed, // Carpet
     0xde, 0xe9, 0xfb, // Not scanned space
     0xed, 0xf3, 0xfb, // Possible obstacle
+    0xa2, 0xbc, 0xe7, // Room color 1
+    0xec, 0xd0, 0x99, // Room color 2
+    0x9b, 0xd4, 0xda, // Room color 3
+    0xec, 0xc6, 0xc9, // Room color 4
+    0xd7, 0xbc, 0xe3, // Room color 5
+    0xc3, 0xe2, 0xb6, // Room color 6
 ];
-const MAP_IMAGE_PALETTE_LEN: u8 = 6;
+const MAP_IMAGE_PALETTE_LEN: u8 = 12;
 // 0 -> Transparent, 255 -> Fully opaque
 // (first entry is transparent, rest opaque)
 const MAP_IMAGE_PALETTE_TRANSPARENCY: &[u8] = &[0u8, 255, 255, 255, 255, 255];
```
But the maps in the test didn't change so I guess they don't have that added information in them (?). Not sure if we should add that without a map that we can see it working on

src/map/style.rs
shmuelzon: Probably because I don't know any better and rely on compilation errors and LLMs 🤷‍♂️.
The structure of the hash map changed now because of the below request and it swears that it can't be removed anymore.

src/map/style.rs
shmuelzon: I thought there might be a case in the future but I guess we can complicate it in the future, it needed.

src/map/style.rs
shmuelzon: This required changing the structure of the hash map to also have the ID as a separate field since I didn't find a way to get it from the built SVG node and I didn't want to write the ID value twice. That complicated the surrounding code a bit...

edenhaus (approved):
Thanks @shmuelzon :+1:

PR #1305, opened 2025-11-11
Add files via upload

Johan854:


edenhaus:
Please update the PR title to include the name of the model

edenhaus:
Drafting as required changes are not implemented, and why does this N20 model have fewer things than the base N20? See https://github.com/DeebotUniverse/client.py/blob/dev/deebot_client/hardware/kr0277.py

Johan854:
This N20 does not mop; it’s an N20 Combo. The robot only works as a vacuum cleaner, and there is also a stick vacuum. Same N30s combo but not mop

PR #1306, opened 2025-11-11
Improve support for Deebot T50 Max Pro Omni (c8rj4y)

ClassicGOD:
Since custom water amount I was campaigning for was recently added and maps for newer bots got implemented I decided to finally submit more functional hardware file for the Deebot T50 Max Pro Omni as the currently linked file is not compatible with this hardware.

Basically slightly modified n0vyif.py to add round mop support.


edenhaus (approved):
Thanks @ClassicGOD :+1:

Issue #1296, opened 2025-11-08
Add support for DEEBOT T50 OMNI (mezar1)


psym88:
### Checks

- [x] I have read the above text and I'm aware of the situation
- [x] I have searched the existing issues and no issue is describing my issue
- [x] I have installed the latest version

### The feature you want to suggest

First of all, thank you so much for the great work the maintainers have done on this integration!

I just managed to get my Deebot T50 working using a symlink to another Deebot T50 file.

`docker exec -it homeassistant bash`
`cd /usr/local/lib/python3.13/site-packages/deebot_client/hardware`
`ln -svfT fd60kt.py mezar1.py`

No Idea why my Deboot has another Class (mezar1) than the already existing T50 (fd60kt). 

> 2025-11-08 20:27:06.964 WARNING (MainThread) [deebot_client.api_client] Device class "mezar1" not recognized. Please add support for it: {'did': '[REMOVED]', 'class': 'mezar1'

> 2025-11-08 20:27:06.964 WARNING (MainThread) [homeassistant.components.ecovacs.controller] Device "DEEBOT T50 OMNI" not supported. More information at https://github.com/DeebotUniverse/client.py/issues/612: {'did': '[REMOVED]', 'class': 'mezar1'

I'm not familiar with Git, and someone else might appreciate the adaptation. If someone could help with a PR

PR #1308, opened 2025-11-14
Add support for T50 Omni (mezar1)

basvanrooten:
## Goal
Add support for T50 Omni that reports as `mezar1`

##  Why
I've got the T50 Omni, but just like #1296 my Deebot reports as `mezar1` instead of `fd60kt`.

## Proposed fix
Added symlink mezar1.py which points to fd60kt. I tested it locally too and everything seems to work brilliantly.

Hope this is enough, please let me know if additional info is required.


edenhaus (approved):
Thanks @basvanrooten :+1:

PR #1310, opened 2025-11-15
Improve Deebot T30C Gen2 (viq3mw)

MarkusMWolff:
Adds basic implementation for Deebot T30C Gen2. All basic settings should work. The detailed settings are not fully tested. Some settings in the blueprint do not align with the settings on the app. I would appreciate guidance on how to analyze the app's API calls.


dermitschi:
@MarkusMWolff 
This is really great. Your coding did already work for my Deebot T50 Pro Omni Gen2.

Would you mind adding a simlink to your config in xztz07.py?
I could change it on my own but it would be better to have it in your commit/pull request. 
Maybe you can do this and submit pull-request again?

Otherwise I will need to wait until your pull gets merged. 

Thank you in advance!

BR
Michael

edenhaus:
Analysing the API is hard and time-consuming. If you have Apple devices, you can use a man-in-the-middle proxy to analyse the traffic. But you need to install custom certificates and co. Also, the app recently started to add certificate pinning, so it can be tricky.
For my analysis,s I use https://www.mitmproxy.org/

@dermitschi we should not add two different models in the same PR, especially since the T30 and T50 can have different features. By splitting them into two PR it will also help to understand the history in the commits better

dermitschi:
@edenhaus 
thank you for your suggestions.

I tried to manually manipulate the deebot_client files in my HA test environment. 
Now I did exec into my live HA Core and changed it there without success. 

I'am not familiar with python and therefore it is hard for me to completely understand what I need to do. Always trying things but sometimes without success :( 

I will try again and give some feedback. Also I could directly do a commit in my fork and do a pull-request. But testing is a little problem for me :)

Will give my very best. 

BR
Michael

MarkusMWolff:
Hi I did some minor improvements. I hope I also fixed the Conversion from symbolic link to actual file now.

edenhaus (approved):
Thanks @MarkusMWolff :+1:

PR #1319, opened 2025-11-19
Fix symlink for T50 Pro Omni (elrxgb)

WillBishop:
This replaced the invocations of Clean, CleanArea, and GetCleanInfo with their V2 variants.

Clean does not work on the T50 Pro Omni, CleanV2 does.


edenhaus (approved):
Thanks @WillBishop :+1:

PR #1329, opened 2025-11-22
README: Fix deprecation warning in example code

Rosi2143:
update async start code to remove this warning

DeprecationWarning: There is no current event loop


edenhaus:
Which version of python are you using?

Rosi2143:
Python 3.13.6

PR #1330, opened 2025-11-22
README: Fix missing import of "Charge" command

Rosi2143:
in example code the Charge command was not imported


edenhaus (changes requested):
You are importing the same file multiple times. Please fix it

PR #1457, opened 2026-02-28
Fix room detection for robots with 11+ field subsets

JBG-brad:
## Problem

Robots like the DEEBOT X11 OmniCyclone (model class `huhcip`) return 11-field room subsets from `getMapSet_V2`, but `_handle_rooms_subsets` only matches exactly 10 fields (`len(subsets[0]) == 10`).

This causes the code to fall through to the 8-field path which calls `GetMapSubSet` — a command the X11 doesn't support, returning error code 20003 ("rcp not support"). The result is empty room attributes on vacuum entities in Home Assistant.

## Fix

Change `== 10` to `>= 10`. The 10+ field branch only reads `subset[0]` (id) and `subset[1]` (name), so extra trailing fields are safely ignored.

The 11th field appears to be related to floor material/carpet detection, consistent with the X11's automatic carpet recognition feature.

## Tested on

- DEEBOT X11 OmniCyclone (huhcip), firmware 1.67.0
- deebot-client 17.1.0 installed in Home Assistant OS
- Two vacuums, both now show correct room names after this fix

### DIFF
--- a/deebot_client/commands/json/map/__init__.py
+++ b/deebot_client/commands/json/map/__init__.py
@@ -273,7 +273,7 @@ class GetMapSetV2(GetMapSet):
         map_id: str,
     ) -> HandlingResult:
         # there are two versions of this message, depending on the number of values
-        if subsets and len(subsets[0]) == 10:
+        if subsets and len(subsets[0]) >= 10:
             # subset values
             # 1 -> id
             # 2 -> name
--- a/tests/commands/json/map/test_init.py
+++ b/tests/commands/json/map/test_init.py
@@ -1,9 +1,12 @@
 from __future__ import annotations
 
 from typing import Any
+from unittest.mock import Mock
 
 import pytest
 from testfixtures import LogCapture
+
+from deebot_client.event_bus import EventBus
 
 from deebot_client.commands.json import (
     GetMapSet,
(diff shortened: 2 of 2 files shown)


deebot_client/commands/json/map/__init__.py
```diff
@@ -273,7 +273,7 @@ def _handle_rooms_subsets(
         map_id: str,
     ) -> HandlingResult:
         # there are two versions of this message, depending on the number of values
-        if subsets and len(subsets[0]) == 10:
+        if subsets and len(subsets[0]) >= 10:
```
edenhaus, line 276:
```suggestion
        if subsets and len(subsets[0]) in (10, 11):
```

We should be specific as Ecovacs is constantly changing api
Can you please verify if the values are still matching to your format and add a comment for the 11 field

tests/commands/json/map/test_init.py
```diff
@@ -463,6 +466,66 @@ async def test_getMapSetV2_rooms_v2() -> None:
...
+                    "0",
+                    "0",
+                ],
+            ],
+            [Room("Bathroom", 0, ""), Room("Hallway", 1, "")],
+        ),
+    ],
+    ids=["10-field", "11-field"],
+)
+def test_handle_rooms_subsets_field_count(
```
edenhaus, line 515: We should not test internal functions instead add a test similar to the others

edenhaus (changes requested):

deebot_client/commands/json/map/__init__.py
JBG-brad: I cannot figure out what the 11th field is.  It appears to be binary as I only see 0 or 1 values, but only one room in on each of my vacuums has the value set to `1`.  All of the other rooms are set to `0`. 

Is it okay that I don't know what it's for?

edenhaus: Set it to unknown like the others. Are the other fields the same?

mattenz: Just commenting to add that adding that fix on line 276 enables the functionality in HA 2026.3 where you can map rooms from Ecovacs to Areas in HA for the X11
image

Gw3n4: > I cannot figure out what the 11th field is. It appears to be binary as I only see 0 or 1 values, but only one room in on each of my vacuums has the value set to `1`. All of the other rooms are set to `0`.
> 
> Is it okay that I don't know what it's for?

The room coordinates are composed of 2 digits:
the map number (in my case 1st floor is map 1, 2nd floor is map 2) and the room number.

For instance, in the vavuum card, I define the shortcuts to clean a specific room like this:
- name: Clean dining room
    service: vacuum.send_command
    service_data:
      command: clean_V2
      params:
        act: start
        content:
          type: freeClean
          value: 1,3
    icon: mdi:table-furniture
    target:
      device_id: cd66f34d9c274afd63333669b2aaeca7

Where 1 is the map number,  3 is the room number

I hope this helps.

edenhaus:
Drafting it as requested, changes are not implemented. Set it to ready to review after implementing it

PR #1458, opened 2026-02-28
Add freeClean support for newer robots

JBG-brad:
## Summary

Newer robots like the DEEBOT X11 OmniCyclone ~and X9 PRO OMNI~ use the `freeClean` clean type instead of `spotArea` for room cleaning via `clean_V2`. Without this change, room cleaning commands fail with error 20011 "unknown type".

- Adds `FREE_CLEAN` value to the `CleanMode` enum
- Adds `CleanAreaFreeClean` command class that sends `{"type": "freeClean", "value": "<cleanings>,<room_ids>"}`
- Updates the `huhcip` (ilt3k8) hardware definition to use `CleanAreaFreeClean`
- Adds test cases for the new command class

Follow-up to #1457.

## Testing

Tested on a DEEBOT X11 OmniCyclone — room cleaning commands now work correctly via Home Assistant.

Edit:

Misunderstanding how the hardware definitions work; I assumed the X9 PRO OMNI was the same as the X11.  A new PR is coming to define the X11 OmniCyclone.

### DIFF
--- a/deebot_client/commands/json/clean.py
+++ b/deebot_client/commands/json/clean.py
@@ -108,6 +108,29 @@ class CleanAreaV2(CleanV2):
         return args
 
 
+class CleanAreaFreeClean(CleanV2):
+    """Clean area command using freeClean type.
+
+    Used by newer robots (e.g. DEEBOT X11 OmniCyclone) that use the freeClean
+    clean type instead of spotArea. The value format is "<cleanings>,<room_ids>".
+    """
+
+    def __init__(
+        self, mode: CleanMode, area: list[int | float], cleanings: int = 1
+    ) -> None:
+        self._additional_content = {
+            "type": CleanMode.FREE_CLEAN.value,
+            "value": f"{cleanings},{','.join(str(int(i)) for i in area)}",
+        }
+        super().__init__(CleanAction.START)
+
+    def _get_args(self, action: CleanAction) -> dict[str, Any]:
+        args = super()._get_args(action)
+        if action == CleanAction.START:
+            args["content"].update(self._additional_content)
+        return args
+
+
 class GetCleanInfo(JsonCommandWithMessageHandling, MessageBodyDataDict):
     """Get clean info command."""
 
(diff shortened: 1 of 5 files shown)


edenhaus (commented):
Is freeClean just used for Area/Room cleaning or also for other things?

JBG-brad:
> Is freeClean just used for Area/Room cleaning or also for other things?

As best as I can tell it's just the for area cleaning.  

Theres some nonsensical (I haven't deciphered it yet) data associated with freeClean that I think is to send the robot to clean only a carpet in the room.

shmuelzon:
Hey @JBG-brad, I was just looking into this since HA added the `clean_area` action and I see the same is also true for my X9 Pro (`lwmdoj`) but I believe it'll be the same for the X9 Omni too (`ilt3k8`).
Is there a reason why you reverted the change that fixed this functionality there too and opted to create a new file for the X11? If it's just that you didn't know if it's relevant for those model then I can confirm that it is.

My app sends the same `clean_V2` when cleaning two rooms (where the room IDs are 15 and 12) with:
```json
{
    "data": {
        "act": "start",
        "content": {
            "type": "freeClean",
            "value": "1,15;1,12"
        }
    }
}
```

The only thing that doesn't seem to match from your PR is the `value`. Where you wrote:
```python
"value": f"{cleanings},{','.join(str(int(i)) for i in area)}",
```
I would have expected something like the following:
```python
"value": ";".join(f"{cleanings},{int(i)}" for i in area),
```
(Not even sure what the `cleanings`/`1` value really is)

deebot_client/commands/json/clean.py
```diff
@@ -108,6 +108,29 @@ def _get_args(self, action: CleanAction) -> dict[str, Any]:
         return args
 
 
+class CleanAreaFreeClean(CleanV2):
```
edenhaus, line 111: Please merge this command into the clean area above as the only difference is, how the value is specified.
So create a if statement to change it in case of free clean

edenhaus (changes requested):

deebot_client/commands/json/clean.py
N1KHL: @edenhaus I used the requested approach on #1607

edenhaus:
Already implemented

PR #1473, opened 2026-03-20
Add Deebot X5 PRO OMNI support

genelee26:
Add support for DEEBOT X5 PRO OMNI, merge the "CapabilityWater" from X2 and base on the X8 Pro.


edenhaus:
Did you test this change?

genelee26:
Sorry for the late reply — I was on vacation and just got back to this.

Yes, I tested all of this on my physical Deebot X5 PRO OMNI device and verified sensor readings against the official Ecovacs app.

**Testing process:**

1. Started with the X8 PRO definition as the base, since the X5 PRO OMNI shares most capabilities with it. Ran a full test of all capabilities — clean, charge, fan speed, map, life span, station, settings, stats, etc. All readings matched the app.

2. During testing, I found that `water_amount` on the X5 PRO OMNI uses discrete levels (LOW/MEDIUM/HIGH) rather than a continuous slider like the X8 PRO. The level count matched the X2's implementation exactly.

3. Merged `CapabilityWater` from the X2 definition, then re-tested the water module independently — water amount switching and mop_attached status both worked correctly and matched the app.

All capabilities defined in this PR have been verified on real hardware.

edenhaus (approved):
Thanks @genelee26 👍

PR #1475, opened 2026-03-20
Fix mowing actions for GOAT A1600 RTK

reniko:
PR is co-authored with Copilot (Gemini 3.1).

I found existing support for Evovacs GOAT A1600 RTK (xmp9ds) but mowing actions failed in home assistant.

To fix the folling was implemented:
* Introduced the `CleanMower` class, a variant of `CleanV2` tailored for mower devices that use the "clean" command name with the V2 content format.

Fix was manually copied to my home assistant and was sucessfully testet. GOAT started mowing and returned to dock.

Furthermore Gemini did some enhancements of the test suite which I wasn't able to test. But from my point of view they look promissing ;-)


deebot_client/commands/json/clean.py
```diff
@@ -84,13 +84,19 @@ def _get_args(self, action: CleanAction) -> dict[str, Any]:
...
                 content["type"] = CleanMode.AUTO.value
             case CleanAction.STOP | CleanAction.PAUSE:
                 content["type"] = ""
         return args
 
 
+class CleanMower(CleanV2):
+    """Clean command for mower devices that use the 'clean' command name with V2 content format."""
+
+    NAME = "clean"
```
edenhaus, line 97: We already have a Clean command. See above.

Can you test if that one is working?
I don't want to create two commands with the same name

reniko: Thanks for the review!

I tried my best to understand the way everything is working. After some sessions with Github Copilot (Claude Sonnet 4.6 this case):

Clean and CleanMower cannot substitute each other — they send structurally different payloads for the same API command name:

```
Clean(CleanAction.START)._args        # {"act": "start", "type": "auto"}
CleanMower(CleanAction.START)._args   # {"act": "start", "content": {"type": "auto"}}
```

The GOAT mower API requires the nested content dict (the V2 content format). If Clean is used instead, "type": "auto" is sent at the top level, which the device does not recognise — this is exactly the root cause of the original mowing failure that was tried to fix.

reniko:
Two additional changes have been included since the initial version:

**1. Remove incorrect test_CleanMower_args (tests/commands/json/test_clean.py)**

The AI-generated test_CleanMower_args test had wrong expectations: it asserted content: {"type": "auto"} for PAUSE and STOP, but the implementation (inherited from CleanV2._get_args) correctly sets content: {"type": ""} for those actions. The test was also fully redundant with the already-correct test_CleanMower_content test that was present just above it. It has been removed.

(in the end one Claude Sonnet 4.6 fixed something from Gemini 3.1 Pro)

**2. Fix GetMapTrace parsing for the v2 format (deebot_client/commands/json/map/__init__.py)**

After the inital fix there were tons of errors regarding "Could not parse getMapTrace"

GOAT mower devices deliver map trace data in a different format than vacuum robots. Instead of the v1 payload {totalCount, traceStart, traceValue}, they send a v2 payload {mid, batid, serial, index, type, info, infoSize}. The GetMapTrace handler now detects this v2 format by checking for the presence of the "info" key and extracts the trace data accordingly — note that index is 1-based in the v2 format and is converted to a 0-based start offset. A new parametrized test test_getMapTrace_v2 covering index values 1 → 3 has been added to verify this path.

reniko:
I hope we're coming closer to a merge of the PR. It's running with that config since ~ a week now with daily mowing started from home assistant.

I wondered if it could make sense to implement a map for mowers since we already have the position data available. Maybe you could point me in the right direction? I failed while trying to debug the MQTT responses from the mower.

deebot_client/commands/json/clean.py
edenhaus: Thanks for the explanation.

The only issue I currently have with the proposed solution is that we use the command name as a lookup in 
https://github.com/DeebotUniverse/client.py/blob/cc7bef44e4e925703f0fac33f9a256f4ae4a68d9/deebot_client/commands/json/__init__.py#L261

So, currently, your command would replace the other clean command. This lookup is not used for the clean command but we should fix it first.
The issue is that we currently do a name lookup, and the code is not prepared to differentiate between two commands as the context is missing.
I currently don't know how to fix it fast so input is welcome

edenhaus:
This PR cannot be merged currently due https://github.com/DeebotUniverse/client.py/pull/1475#discussion_r3005236742

For the map. Please see how bots are using it and try to configure for your bot. Please be aware there are mutliple version of the map and probably you should try the second version first

edenhaus:
What are the claude workflows???

stephanerosi:
Hello @reniko 
Any reason why you closed the pr?

reniko:
Hi,

replaced with #1515.

PR #1507, opened 2026-03-30
Fix clean area v2 for HA

MarkusMWolff:
Makes the clean area method available to Home Assistant for newer models using the V2 API. 
Related to: https://github.com/home-assistant/core/issues/165933

### DIFF
--- a/deebot_client/commands/json/clean.py
+++ b/deebot_client/commands/json/clean.py
@@ -95,8 +95,9 @@ class CleanAreaV2(CleanV2):
     """Clean area command."""
 
     def __init__(self, mode: CleanMode, area: list[int | float], _: int = 1) -> None:
+        mode = CleanMode.FREE_CLEAN
         self._additional_content = {
-            "type": CleanMode.FREE_CLEAN.value,
+            "type": mode.value,
             "value": ";".join(("1," + str(i)) for i in area),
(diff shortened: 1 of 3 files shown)


deebot_client/commands/json/clean.py
```diff
@@ -95,10 +95,16 @@ class CleanAreaV2(CleanV2):
     """Clean area command."""
 
     def __init__(self, mode: CleanMode, area: list[int | float], _: int = 1) -> None:
-        self._additional_content = {
-            "type": mode.value,
-            "value": ",".join(str(i) for i in area),
-        }
+        if mode == CleanMode.SPOT_AREA:
```
edenhaus, line 98: This will break many bots as for example my bot does not support freeclean.

We don't can change it like this... We need to extend the capabilities so this command is aware of the model and changes only for the models which need it

edenhaus (changes requested):

Gr4vlty:
Tested on T50 Pro Omni Gen2 (xztz07). The correct value format for freeClean is "count,room_id" per room, joined by semicolons. Example for two rooms: "1,1;1,2". Using "spotArea" as type returns code 20011. The fix should format value as ";".join(f"1,{int(i)}" for i in area).

PR #1515, opened 2026-04-02
Add CleanMower command for mower devices

reniko:
## Summary
Introduces a new `CleanMower` command class to support mower devices that use the "clean" endpoint with V2 content format, while refactoring the shared V2 argument logic.

## Key Changes
- **New `CleanMower` class**: A specialized clean command for mower devices that uses the "clean" endpoint (not "clean_V2") but with V2-style content formatting
- **Refactored V2 argument logic**: Extracted the V2 content format logic from `CleanV2._get_args()` into the base `Clean._get_args()` method, controlled by a `_v2_args` class variable
  - `Clean._v2_args = False` (default V1 format)
  - `CleanV2._v2_args = True` (V2 format)
  - `CleanMower._v2_args = True` (V2 format)
- **Updated hardware configs**: Changed xmp9ds and 300lc5 device configs to use `CleanMower` instead of `CleanV2`
- **Updated xmp9ds device info**: Changed from `GetCleanInfoV2()` to `GetCleanInfo()` for state events
- **Added comprehensive tests**: New test cases for `CleanMower` content formatting and `CleanV2` content formatting to ensure correct behavior across all action types (START, RESUME, PAUSE, STOP)

## Implementation Details
The V2 content format follows this pattern:
- START/RESUME actions: `{"type": "auto"}`
- PAUSE/STOP actions: `{"type": ""}`

This refactoring eliminates code duplication while maintaining backward compatibility with existing `Clean` and `CleanV2` commands.

https://claude.ai/code/session_01YFYjxwixRZrtjfv1aUfoVQ

### DIFF
--- a/deebot_client/commands/json/__init__.py
+++ b/deebot_client/commands/json/__init__.py
@@ -13,7 +13,7 @@ from .carpet import GetCarpetAutoFanBoost, SetCarpetAutoFanBoost
 from .charge import Charge
 from .charge_state import GetChargeState
 from .child_lock import GetChildLock, SetChildLock
-from .clean import Clean, CleanArea, CleanV2, GetCleanInfo, GetCleanInfoV2
+from .clean import Clean, CleanArea, CleanMower, CleanV2, GetCleanInfo, GetCleanInfoV2
 from .clean_count import GetCleanCount, SetCleanCount
 from .clean_logs import GetCleanLogs
 from .clean_preference import GetCleanPreference, SetCleanPreference
@@ -62,6 +62,7 @@ __all__ = [
     "Charge",
     "Clean",
     "CleanArea",
+    "CleanMower",
     "CleanV2",
     "ClearMap",
     "GetAdvancedMode",
--- a/deebot_client/commands/json/clean.py
+++ b/deebot_client/commands/json/clean.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any, ClassVar
+from weakref import WeakKeyDictionary
 
 from deebot_client.events import StateEvent
 from deebot_client.logging_filter import get_logger
@@ -23,6 +24,7 @@ class Clean(ExecuteCommand):
 
     NAME = "clean"
     _v2_args: ClassVar[bool] = False
+    _active_clean_type: ClassVar[WeakKeyDictionary] = WeakKeyDictionary()
 
     def __init__(self, action: CleanAction) -> None:
         super().__init__(self._get_args(action))
@@ -47,13 +49,23 @@ class Clean(ExecuteCommand):
             ):
                 self._args = self._get_args(CleanAction.RESUME)
 
+            # V2 commands: match the active job's clean type on pause/resume/stop
+            content = self._args.get("content")
+            if isinstance(content, dict) and self._args["act"] in (
+                CleanAction.PAUSE.value,
+                CleanAction.RESUME.value,
+                CleanAction.STOP.value,
+            ):
+                content["type"] = self._active_clean_type.get(
+                    event_bus, CleanMode.AUTO.value
+                )
+
         return await super()._execute(authenticator, device_info, event_bus)
 
     def _get_args(self, action: CleanAction) -> dict[str, Any]:
         if self._v2_args:
             content: dict[str, str] = {}
             args: dict[str, Any] = {"act": action.value, "content": content}
-            # Cloud rejects empty type with error 20003; real app always sends a mode
             content["type"] = CleanMode.AUTO.value
             return args
         args = {"act": action.value}
@@ -143,6 +155,9 @@ class GetCleanInfo(JsonCommandWithMessageHandling, MessageBodyDataDict):
             content = clean_state.get("content", {})
             if "type" in content:
                 clean_type = content.get("type")
+
+            if clean_type:
+                Clean._active_clean_type[event_bus] = clean_type
 
             if clean_type == "customArea":
                 area_values = content
--- a/deebot_client/commands/json/clean.py
+++ b/deebot_client/commands/json/clean.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, ClassVar
 
 from deebot_client.events import StateEvent
 from deebot_client.logging_filter import get_logger
@@ -22,6 +22,7 @@ class Clean(ExecuteCommand):
     """Clean command."""
 
     NAME = "clean"
+    _v2_args: ClassVar[bool] = False
 
     def __init__(self, action: CleanAction) -> None:
         super().__init__(self._get_args(action))
@@ -49,6 +50,15 @@ class Clean(ExecuteCommand):
         return await super()._execute(authenticator, device_info, event_bus)
 
     def _get_args(self, action: CleanAction) -> dict[str, Any]:
+        if self._v2_args:
+            content: dict[str, str] = {}
+            args: dict[str, Any] = {"act": action.value, "content": content}
+            match action:
+                case CleanAction.START | CleanAction.RESUME:
+                    content["type"] = CleanMode.AUTO.value
+                case CleanAction.STOP | CleanAction.PAUSE:
+                    content["type"] = ""
+            return args
         args = {"act": action.value}
         if action == CleanAction.START:
             args["type"] = CleanMode.AUTO.value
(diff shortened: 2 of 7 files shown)


reniko:
Ok, change is running good now since sone days. Is that implementation now as you requested in my other change?

Sorry again for some unplanned pushes to your repo. I still struggle when GitHub merges something upstream when I work in a fork.

Anything that should be enhanced or changed?

reniko:
Anything that could be done from my side for this PR (except waiting for the review)?

shinerblue:
Testing this PR's `CleanMower` approach on a GOAT A3000 LiDAR Pro (`51rcxt`, newly added in #1538) — found that the current `{"type": ""}` payload for `PAUSE` and `STOP` is rejected by the Ecovacs cloud:

```
WARNING [deebot_client.commands.json.common] Command "clean" was not successfully. body={'code': 20003, 'msg': 'unknow type'}
```

Start works, pause/stop fail with that 20003 error.

Cross-referencing the iOS-app traffic capture in [issue #852, comment 2745410917](https://github.com/DeebotUniverse/client.py/issues/852#issuecomment-2745410917):

```
pause: {"body":{"data":{"act":"pause","content":{"type":"auto"}}}}
stop:  {"body":{"data":{"act":"stop","content":{"type":"auto"}}}}
```

The real app sends `{"type": "auto"}` for **all four** actions — `start` / `resume` / `pause` / `stop` — not empty string. The Ecovacs backend appears to treat the empty string as an unknown clean type and rejects the command.

One-line fix on top of this PR — in `Clean._get_args`, the V2 branch should set `content["type"] = CleanMode.AUTO.value` unconditionally for the four clean actions, not only for `START | RESUME`:

```python
if self._v2_args:
    content: dict[str, str] = {}
    args: dict[str, Any] = {"act": action.value, "content": content}
    content["type"] = CleanMode.AUTO.value
    return args
```

Verified on real hardware (GOAT A3000 LiDAR Pro, HAOS 17.2, HA core-2026.4.3, `deebot-client` 18.1.0 with this PR's changes + the one-line correction):

```
CleanMower(START)  -> {'act': 'start',  'content': {'type': 'auto'}}
CleanMower(RESUME) -> {'act': 'resume', 'content': {'type': 'auto'}}
CleanMower(PAUSE)  -> {'act': 'pause',  'content': {'type': 'auto'}}
CleanMower(STOP)   -> {'act': 'stop',   'content': {'type': 'auto'}}
```

Pause, resume from paused, start, and dock all work end-to-end from the HA lawn_mower card — no more 20003 from the cloud, mower state reliably transitions to `paused` and back.

Happy to open a follow-up PR on top of this one once it merges. Separately, `51rcxt.py` (just landed in #1538) also needs the `CleanMower` + `GetCleanInfo` swap — I can include that in the follow-up.

@reniko @edenhaus — let me know if you'd prefer I push the fix into this PR or keep it separate.

shinerblue:
Heads up — I captured the iOS Ecovacs app's actual MQTT payloads for start/pause/resume/stop on a GOAT A3000 (class `51rcxt`, firmware `1.7.20`) and the rule is simpler than the current `CleanMower` approach in this PR assumes. Full write-up with 10-row capture table and example frames here: [#852 (comment)](https://github.com/DeebotUniverse/client.py/issues/852#issuecomment-4315741963).

Short version:

```
act     | content
--------|----------------------------------------
start   | {type: <mode>, value: <id>?}     ← value only for spotArea/borderrotate, never for auto
pause   | {type: <active_type>}            ← value never sent
resume  | {type: <active_type>}            ← value never sent
stop    | {type: <active_type>}            ← value never sent
```

Two things this suggests for the PR:

1. **Pause/resume/stop should send `content: {type: <cached_active_type>}`, no `value`** — the iOS app never sends `value` on those verbs, even when the active job started with one. The mower accepts `value` if sent (our local patch over-sends and it works), but matching the app's minimal payload is cleaner.
2. **Empty `type` ("") on pause/resume is what gets rejected with `code 20003 "unknow type"`** — which is consistent with what you'd see using `CleanV2`'s default. Fixing the empty-`type` case alone isn't enough, though — sending `{type: "auto"}` for every pause/resume silently fails when the active job is `spotArea` (cloud accepts with `code: 0` but the mower doesn't react). You need the cached-`type` echo.

Happy to review or contribute once #1515 settles. Our soft-fork (`shinerblue/client.py @ a3000-cleanmower-autofix`) is currently using a WeakKeyDictionary on the event bus to hold the cached type; we populate it from the `onCleanInfo` MQTT snoop. I can open a follow-up PR with the simpler-than-we-had-it approach on top of whatever lands here.

Rico36:
**Confirmed fix needed for GOAT A3000 LiDAR (`cr0e4u`, fw 1.13.31)**

Manually applied PR 1515 proposed `GetCleanInfoV2` → `GetCleanInfo` fix to `cr0e4u.py` on my live Home Assistant (HA) Container instance.  Prior to the patch, `getCleanInfo_V2` was consistently returning `errno 500 / wait for response timed out` (~20s) on every poll cycle. The device simply doesn't implement the V2 endpoint and `StateEvent` never transitioned out of `DOCKED`.

Post-patch with `GetCleanInfo`: device ACKs in ~176ms confirming the endpoint works. However, further testing revealed `getCleanInfo` consistently returns `state: idle` on this device regardless of actual mowing state and making it unreliable as a state (Docked vs Mowing vs Paused, etc) source. 

FOR CONSIDERATION:
The real-time state update mechanism on this device is `onScheduleTaskInfo`, pushed at session start with the authoritative state:
```json
{"state": "clean", "cleanState": {"motionState": "working", "content": {"type": "spotArea", "value": "3,2,6,4,7,5"}}}
```
Even with PR 1515 this message is currently dropped as `Unknown message "onScheduleTaskInfo"`.  A handler or message registration for `onScheduleTaskInfo` is definitely needed.  Fortunately, the payload maps directly to existing `GetCleanInfo._handle_body_data_dict` state logic. One possible follow-up enhancement would be to introduce a GetScheduleTaskInfo command class and evaluate whether it provides a more accurate polling source for GOAT mower state than getCleanInfo.

One thing worth noting:  while this PR targets xmp9ds and 300lc5, the same GetCleanInfoV2 → GetCleanInfo fix is equally needed for **cr0e4u.py** (GOAT A3000 LiDAR) and **5xu9h3.py** (GOAT G1), which are not currently covered in this PR.  Both files independently declare `GetCleanInfoV2` in their state capability.

Please do get this merged — and the above findings for `cr0e4u` are offered as input for follow-on work.

---
Thanks,
-Ricky

Rico36:
Following up on my earlier comments: I’ve put everything together in a public repo in case it’s useful as reference material or for testing against real A3000 hardware.
https://github.com/Rico36/Ecovacs-Goat-A3000-Mower
It includes the three patched files (cr0e4u.py, clean.py, messages/json/init.py), the apply/check scripts, and a findings log covering confirmed payloads, state routing, and error codes on fw 1.13.31. Happy to test any branch from this PR directly on the device.

mbrackjr:
Independently verified this same approach on a different GOAT model — a GOAT A1600 LiDAR Pro (e4gqia), not the model this PR was originally built against.

Setup: deebot-client 18.5.1, Home Assistant 2026.8.1, HAOS on Raspberry Pi 4B.

e4gqia.py was previously wired to CleanV2/GetCleanInfoV2 (added generically via #1554's similarity matching, untested against real hardware). Swapping it to the CleanMower approach from this PR — clean endpoint, {"type": "auto"} — fixed start/pause/stop, which were non-functional before.

Hope someone with coding skills can get the conflicts resolved so my robot becomes properly supported.

Patch76:
**GOAT O1200 LiDAR (`2i0fns`) — not yet covered here, plus the untested `borderrotate` half**

Adding a model that this PR does not touch: on `dev`, `2i0fns` still binds `CleanV2` and `GetCleanInfoV2`, and neither it nor `0jbd6s` imports `CleanAreaV2` — both lack the `area=` argument that `300lc5` and `xmp9ds` carry on line 84.

**Two distinguishable failure modes.** On this device three `clean_V2` calls (one `pause`, two `start`) and every `getCleanInfo_V2` poll timed out after 20 s with `{'ret':'fail','errno':500,'debug':'wait for response timed out'}` — no rejection, no answer at all. That is a different signature from the `code 20003` @shinerblue reported: 20003 means the command name was accepted and the payload rejected, while a silent timeout points at the name itself. In the same session `getBattery`, `getChargeState`, `charge`, `getAreaParameter`, `setAreaParameter` and `getCleanInfo` (no `_V2`) all answered `code 0` within a second, so the device stayed reachable throughout.

The app's request for the same job differs only in the command name:

```
HA   cmdName clean_V2  body.data {"act":"start","content":{"type":"auto"}}  -> timeout
app  iot/p2p/clean/.../q/  {"act":"start","content":{"type":"auto"}}        -> {"code":0,"msg":"ok"}
```

**The `borderrotate` half you left open.** In #852 you documented the start payload for trimming and noted you hadn't re-tested pause/resume for that mode. I captured the stop on this model, and it follows your rule exactly:

```
start  {"act":"start","content":{"type":"borderrotate","value":"reid:2;reid:3;reid:1"}}
stop   {"act":"stop","content":{"type":"borderrotate"}}
```

`type` alone on stop, no `value`, even though the job carried one at start. The device echoed the same list back in `onCleanInfo`.

I am running `clean` with `{"type":"auto"}` locally for start and stop; each was exercised once and behaved as expected. One separate observation: `charge {"act":"go"}` alone does **not** end a job — afterwards the device still reports `state: clean`, and only `{"act":"stop"}` returns it to `idle`.

**Request:** could `2i0fns` be switched over as well — and `0jbd6s`, whose module docstring is a verbatim copy of the `2i0fns` one, class ID included?

Device: GOAT O1200 LiDAR Pro, class `2i0fns`, firmware 1.11.31, EU region.

PR #1523, opened 2026-04-07
Add support for N50 Pro Omni (a2ywac)

ankitduseja:
DEEBOT N50 PRO OMNI Capabilities added


edenhaus (approved):
Thanks @ankitduseja 👍

Issue #1376, opened 2025-12-22
Disable getMapTrace() for Ecovacs Goat Lawn Mower


robinhood-code:
### Checks

- [x] I have read the above text and I'm aware of the situation
- [x] I have searched the existing issues and no issue is describing my issue
- [x] I have installed the latest version

### The feature you want to suggest

Whenever my lawn mower robot runs, my HA will be flooded with the following "could not parse" message:

As we simply can't display the map data in HA for the lawn mower, can we just disable the parsing of map data for lawn mower? 

```
Logger: deebot_client.message
Source: runner.py:289
First occurred: 5:08:27 PM (5766 occurrences)
Last logged: 6:18:33 PM

[...]

PR #1564, opened 2026-04-29
test(hardware): cover xmp9ds (GOAT A1600 RTK) in capabilities event extraction test

Beennnn:
## Summary

The hardware definition for `xmp9ds` (Ecovacs GOAT A1600 RTK) is already present in the repo and identical in capability set to `5xu9h3` (GOAT G1) and `aadham` (GOAT A3000 LiDAR Pro). However, only `test_all_models_loaded` exercises it — that test only confirms the module loads, not which event refresh commands are exposed.

This PR adds an explicit `xmp9ds` parametrize entry to `test_capabilities_event_extraction` mirroring the existing `5xu9h3` assertions. It locks in the contract that GOAT A1600 RTK has the same capability surface as the other GOATs.

## Why

A user (running firmware 1.15.13) reported in HA logs hundreds of thousands of `Could not parse getMapTrace` warnings per day. Investigation showed that `xmp9ds.py` exists upstream so the model IS recognized. The map trace warnings are caused separately by the legacy fallback in `messages/json/__init__.py:_LEGACY_USE_GET_COMMAND` trying to parse spontaneous `getMapTrace` pushes from the firmware, even though MOWER hardware definitions don't expose a `map=` capability. That is being addressed in a follow-up PR.

In the meantime, having explicit test coverage for `xmp9ds` would have made the diagnostic easier — the missing test entry initially led me to assume `xmp9ds.py` was missing entirely. Adding the assertion locks the model down properly.

## Refs

- DeebotUniverse/client.py#852 (GOAT A1600 RTK Support)
- DeebotUniverse/client.py#1376 (Disable getMapTrace for GOAT — separate follow-up PR)

## Test plan

- [x] `uv run pytest tests/hardware/test_init.py -v` — all 8 tests pass including new `test_capabilities_event_extraction[xmp9ds]`
- [x] No new file added — pure test addition, no production code changed

### DIFF
--- a/tests/hardware/test_init.py
+++ b/tests/hardware/test_init.py
@@ -142,6 +142,30 @@ async def test_get_static_device_info(
             },
         ),
         (
+            "xmp9ds",
+            {
+                AdvancedModeEvent: [GetAdvancedMode()],
+                AvailabilityEvent: [GetBattery(is_available_check=True)],
+                BatteryEvent: [GetBattery()],
+                BorderSwitchEvent: [GetBorderSwitch()],
+                CutDirectionEvent: [GetCutDirection()],
+                ChildLockEvent: [GetChildLock()],
+                CrossMapBorderWarningEvent: [GetCrossMapBorderWarning()],
+                CustomCommandEvent: [],
+                ErrorEvent: [GetError()],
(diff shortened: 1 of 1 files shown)


edenhaus:
Closing as this PR is completely Ai driven and makes no sense. It's just duplicating a test....

PR #1565, opened 2026-04-29
fix(messages): skip map-related legacy fallback for mower devices

Beennnn:
## Summary

When a mower (Ecovacs GOAT G1, A1600 RTK, A3000 LiDAR Pro, …) is mowing, its firmware pushes spontaneous `onMapTrace` MQTT messages. They flow through `messages.get_message()` which falls back via `_LEGACY_USE_GET_COMMAND` to `GetMapTrace`, whose parser does not support the on-wire format used by recent mower firmware (observed: `1.15.13`). The parse fails inside `_handle_error_or_analyse` and emits a `WARNING` per push.

In one user setup this produced **217 520 `Could not parse getMapTrace` warnings over 3 days** (~70k/day, ~50/min during mowing), bloating the HA `homeassistant.log` and the recorder DB by ~100 MB.

Mowers do not expose a `map=` capability in their hardware definitions (see `0jbd6s`, `2ap5uq`, `2i0fns`, `5xu9h3`, `aadham`, `xmp9ds`, `itk04l`), so even if the legacy fallback succeeded, the parsed map-trace data would have nowhere to land. The fallback is pure noise on these devices.

## Change

Add an optional `device_type: DeviceType | None = None` argument to `get_message` and `get_legacy_message`. When the device is a `MOWER`, map-related legacy messages are skipped silently before reaching the parser:

```python
_MAP_LEGACY_COMMANDS = frozenset({
    "getCachedMapInfo", "getMapSet", "getMapSubSet",
    "getMapTrace", "getMinorMap", "getMultiMapState",
})
```

Vacuums (`DeviceType.VACUUM`) and callers without a specified type (`None`) fall through unchanged — fully backwards compatible.

`Device._handle_message` reads `device_type` defensively via `getattr` so existing tests mocking `Capabilities` with `spec_set=` (which excludes dataclass field names from `dir()`) continue to work — they fall back to the no-filter path.

## Refs

- Closes DeebotUniverse/client.py#1376 — Disable `getMapTrace()` for Ecovacs Goat Lawn Mower
- Related DeebotUniverse/client.py#852 — Ecovacs GOAT A1600 RTK Support

## Test plan

- [x] `test_get_messages` (existing 7 cases) — pass unchanged
- [x] `test_get_messages_device_type_filter` (10 new cases) — verify mowers skip map-related legacy fallbacks while non-map messages and other device types remain unaffected
- [x] Full suite: **707/707 tests pass**, no regression
[...]

PR #1566, opened 2026-04-29
fix(message): rate-limit "Could not parse" warnings per NAME

Beennnn:
## Summary

When a firmware version pushes messages whose on-wire format the lib does not yet support, every push triggers a `Could not parse <NAME>` WARNING via `_handle_error_or_analyse`. In one observed case a mower's spontaneous `getMapTrace` pushes produced **217 520 identical entries in 3 days** (~70k/day, ~50/min during mowing), bloating the parent process' log file and any downstream sink (HA recorder DB, Loki, etc.).

The first few warnings are useful — they let an operator notice that something is wrong. Beyond that, identical repeats add no information and only make the rest of the log harder to read.

## Change

Per-NAME counter with a default threshold of 3:
- First 3 parse failures for a given message NAME → WARNING (as before)
- 3rd warning is followed by a single notice: *"Further 'Could not parse <NAME>' entries will be logged at DEBUG level"*
- 4th and beyond → DEBUG

Counters are isolated per NAME, so a **new** message class hitting a parse error still surfaces normally even if another NAME has been silenced.

The 3 existing emission sites all route through the new `_log_parse_failure` helper:
- `_handle_error_or_analyse` exception branch (line 68)
- `_handle_error_or_analyse` ERROR-state branch (line 84)
- `MessageBodyData.__handle_body_data` exception branch (line 259)

Behaviour for non-throttled flows (`SUCCESS`, `ANALYSE`) is unchanged.

## Refs

- Related to DeebotUniverse/client.py#1376
- Companion to DeebotUniverse/client.py#1565 (the `getMapTrace` mower-skip — together they should fully eliminate the user's observed log storm)

## Test plan

- [x] `test_warn_once_throttles_repeated_parse_failures` — calls `Message.handle()` THRESHOLD + 2 times on a class whose `_handle` always raises, asserts THRESHOLD + 1 WARNINGs (incl. the "switching to DEBUG" notice) followed by 2 DEBUGs
- [x] `test_warn_once_isolated_per_message_name` — burns through the threshold for one NAME, verifies a sibling NAME still WARNs on its first parse failure
- [x] Full suite: **699/699 tests pass**, no regression

## Notes for review

- The threshold (3) is module-level; happy to make it configurable via env var or `Message` class attribute if that fits the project style better.
- Counter is module-level state; not reset across the lifetime of a process. A "Restart the parent process to reset" note is included in the threshold-reached message. If you'd rather have an admin API to reset, happy to add it.

### DIFF
--- a/deebot_client/message.py
+++ b/deebot_client/message.py
@@ -29,9 +29,7 @@ _PARSE_FAILURE_THRESHOLD = 3
 _parse_failure_counts: dict[str, int] = {}
 
 
-def _log_parse_failure(
-    name: str, data: object, *, exc_info: bool = False
-) -> None:
+def _log_parse_failure(name: str, data: object, *, exc_info: bool = False) -> None:
     count = _parse_failure_counts.get(name, 0) + 1
     _parse_failure_counts[name] = count
     if count <= _PARSE_FAILURE_THRESHOLD:
--- a/deebot_client/message.py
+++ b/deebot_client/message.py
@@ -25,13 +25,6 @@ _LOGGER = get_logger(__name__)
 
 MessagePayloadType = str | bytes | bytearray | dict[str, Any]
 
-
-# When the on-wire format of a message diverges from what the lib expects
-# (e.g. a firmware bumps an envelope schema), every push from the device
-# triggers a "Could not parse" warning. In one observed case this produced
-# 217 520 identical entries in 3 days. Cap the warnings per message NAME
-# and downgrade subsequent occurrences to DEBUG so a future genuine
-# parse error in *another* NAME still surfaces.
 _PARSE_FAILURE_THRESHOLD = 3
 _parse_failure_counts: dict[str, int] = {}
 
@@ -39,18 +32,14 @@ _parse_failure_counts: dict[str, int] = {}
 def _log_parse_failure(
     name: str, data: object, *, exc_info: bool = False
 ) -> None:
-    """Log a "Could not parse" entry, downgrading to DEBUG past a threshold per NAME."""
     count = _parse_failure_counts.get(name, 0) + 1
     _parse_failure_counts[name] = count
     if count <= _PARSE_FAILURE_THRESHOLD:
         _LOGGER.warning("Could not parse %s: %s", name, data, exc_info=exc_info)
         if count == _PARSE_FAILURE_THRESHOLD:
             _LOGGER.warning(
-                "Further 'Could not parse %s' entries will be logged at DEBUG level"
-                " (reached %d occurrences). Restart Home Assistant or the parent"
-                " process to reset the counter.",
+                "Further 'Could not parse %s' entries will be logged at DEBUG level",
                 name,
-                _PARSE_FAILURE_THRESHOLD,
             )
     else:
         _LOGGER.debug("Could not parse %s: %s", name, data, exc_info=exc_info)
--- a/deebot_client/message.py
+++ b/deebot_client/message.py
@@ -26,6 +26,36 @@ _LOGGER = get_logger(__name__)
 MessagePayloadType = str | bytes | bytearray | dict[str, Any]
 
 
+# When the on-wire format of a message diverges from what the lib expects
+# (e.g. a firmware bumps an envelope schema), every push from the device
+# triggers a "Could not parse" warning. In one observed case this produced
+# 217 520 identical entries in 3 days. Cap the warnings per message NAME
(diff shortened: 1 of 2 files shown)


deebot_client/message.py
```diff
@@ -26,6 +26,36 @@
 MessagePayloadType = str | bytes | bytearray | dict[str, Any]
 
 
+# When the on-wire format of a message diverges from what the lib expects
+# (e.g. a firmware bumps an envelope schema), every push from the device
+# triggers a "Could not parse" warning. In one observed case this produced
+# 217 520 identical entries in 3 days. Cap the warnings per message NAME
+# and downgrade subsequent occurrences to DEBUG so a future genuine
+# parse error in *another* NAME still surfaces.
```
edenhaus:
```suggestion
```

deebot_client/message.py
```diff
@@ -26,6 +26,36 @@
...
+    """Log a "Could not parse" entry, downgrading to DEBUG past a threshold per NAME."""
+    count = _parse_failure_counts.get(name, 0) + 1
+    _parse_failure_counts[name] = count
+    if count <= _PARSE_FAILURE_THRESHOLD:
+        _LOGGER.warning("Could not parse %s: %s", name, data, exc_info=exc_info)
+        if count == _PARSE_FAILURE_THRESHOLD:
+            _LOGGER.warning(
+                "Further 'Could not parse %s' entries will be logged at DEBUG level"
+                " (reached %d occurrences). Restart Home Assistant or the parent"
+                " process to reset the counter.",
```
edenhaus:
```suggestion
```

Lib does not know where it is used

tests/test_message.py
```diff
@@ -67,3 +70,91 @@ def test_WronglyImplementedMessage() -> None:
...
+
+    Without this rate-limit, a firmware push storm (e.g. mower map traces with
+    a schema the lib does not know) was observed to produce >200 000 identical
+    WARNING entries in 3 days. The first few are still useful; the rest belong
+    in DEBUG so the rest of HA's log stays readable.
+    """
+    name = _AlwaysRaisingMessage.NAME
+
+    # Reset state so this test is order-independent
+    message_module._parse_failure_counts.pop(name, None)
```
edenhaus: Reset it completely not just for the name

tests/test_message.py
```diff
@@ -67,3 +70,91 @@ def test_WronglyImplementedMessage() -> None:
...
+        raise ValueError("simulated parse failure")
+
+
+def test_warn_once_throttles_repeated_parse_failures(
+    caplog: pytest.LogCaptureFixture,
+) -> None:
+    """After ``_PARSE_FAILURE_THRESHOLD`` warnings, subsequent failures log at DEBUG.
+
+    Without this rate-limit, a firmware push storm (e.g. mower map traces with
+    a schema the lib does not know) was observed to produce >200 000 identical
```
edenhaus: Remove any comment about the warning created in 3 days. I know you use AI for it and it's fine but tell the AI do don't comment it on each functions.... The test don't care about it. Adopt other places too

tests/test_message.py
```diff
@@ -67,3 +70,91 @@ def test_WronglyImplementedMessage() -> None:
...
+
+    by_level: dict[str, list[logging.LogRecord]] = {"WARNING": [], "DEBUG": []}
+    for r in caplog.records:
+        if name in r.getMessage() and r.levelname in by_level:
+            by_level[r.levelname].append(r)
+
+    # First N raw warnings + 1 "switching to DEBUG" notice
+    assert len(by_level["WARNING"]) == _PARSE_FAILURE_THRESHOLD + 1
+    # Then ``extra_calls`` more failures fell through to DEBUG
+    assert len(by_level["DEBUG"]) == extra_calls
```
edenhaus: compare the records directly against the expected list and not the length...
With your setup it could happen that debug entries are before the warning ones, which would be wrong

tests/test_message.py
```diff
@@ -67,3 +70,91 @@ def test_WronglyImplementedMessage() -> None:
...
+
+
+def test_warn_once_isolated_per_message_name(
```
edenhaus, line 150: Adjust this test too

edenhaus (changes requested):

edenhaus (changes requested):
Please make sure CI is passing

edenhaus (approved):
Thanks @Beennnn 👍 

Did some small refactoring in the tests

PR #1588, opened 2026-05-09
feat: add optional position capability for mowers

Beennnn:
## Summary

Adds \`Capabilities.position\` as an optional \`CapabilityEvent[PositionsEvent]\` so devices that don't expose the full map capability (e.g. mowers like the Ecovacs GOAT family) can still report their position to consumers.

Today \`PositionsEvent\` is only fired through \`CapabilityMap.position\`, but mowers don't have a \`map\` capability, so the event never reaches subscribers even though \`getPos\` works fine on the device. Adding a top-level optional \`position\` capability lets mowers wire \`GetPos\` directly without forcing a fake \`map\` capability.

## Changes

- \`deebot_client/capabilities.py\` — \`Capabilities.position: CapabilityEvent[PositionsEvent] | None = None\`
- \`deebot_client/hardware/xmp9ds.py\` — wires \`position=CapabilityEvent(PositionsEvent, [GetPos()])\` on the GOAT A1600 RTK

Other devices stay unaffected (\`position\` defaults to \`None\`; vacuums keep using \`CapabilityMap.position\`).

## Test

\`uv run pytest tests/\` — 703 passed.

### DIFF
--- a/deebot_client/capabilities.py
+++ b/deebot_client/capabilities.py
@@ -276,6 +276,7 @@ class Capabilities(ABC):
     map: CapabilityMap | None = None
     network: CapabilityEvent[NetworkInfoEvent]
     play_sound: CapabilityExecute[[]]
+    position: CapabilityEvent[PositionsEvent] | None = None
     settings: CapabilitySettings
     state: CapabilityEvent[StateEvent]
     station: CapabilityStation | None = None
--- a/deebot_client/hardware/xmp9ds.py
+++ b/deebot_client/hardware/xmp9ds.py
@@ -40,6 +40,7 @@ from deebot_client.commands.json.error import GetError
 from deebot_client.commands.json.life_span import GetLifeSpan, ResetLifeSpan
 from deebot_client.commands.json.network import GetNetInfo
 from deebot_client.commands.json.play_sound import PlaySound
+from deebot_client.commands.json.pos import GetPos
 from deebot_client.commands.json.stats import GetStats, GetTotalStats
 from deebot_client.commands.json.true_detect import GetTrueDetect, SetTrueDetect
 from deebot_client.commands.json.volume import GetVolume, SetVolume
@@ -58,6 +59,7 @@ from deebot_client.events import (
     LifeSpanEvent,
     MoveUpWarningEvent,
     NetworkInfoEvent,
+    PositionsEvent,
     ReportStatsEvent,
     SafeProtectEvent,
     StateEvent,
@@ -102,6 +104,7 @@ def get_device_info() -> StaticDeviceInfo:
             ),
             network=CapabilityEvent(NetworkInfoEvent, [GetNetInfo()]),
             play_sound=CapabilityExecute(PlaySound),
+            position=CapabilityEvent(PositionsEvent, [GetPos()]),
             settings=CapabilitySettings(
                 advanced_mode=CapabilitySetEnable(
                     AdvancedModeEvent, [GetAdvancedMode()], SetAdvancedMode


Beennnn:
Rebased on top of `dev` (was 27 commits behind). No conflicts, suite still 692/692 green locally.

Quick recap:

- Adds an optional `Capabilities.position` (`CapabilityEvent[PositionsEvent]`) so devices without the full `map` capability — mowers in particular — can still surface their position to subscribers.
- Today `PositionsEvent` is only emitted through `CapabilityMap.position`. Mowers have no map capability, so even though `getPos` works on the device, the event never reaches the EventBus.
- Wired into the 4 mower hardware models; vacuum behaviour is unchanged.

Available for any review feedback.

edenhaus:
What would the usecase for it? GetPos gives you back a coordinates that will only help you if you also have a map
The coordinates are just numbers related on the map and not gps coordinates or so

Beennnn:
You're right that `deebotPos`/`chargePos` are internal grid coordinates and not GPS — I don't want to oversell them. Where I think they still earn their place on mowers:

1. **Same coordinate frame as the trace in [#1567](https://github.com/DeebotUniverse/client.py/pull/1567).** The `onMapTrace` payload gives trajectory in the identical `(x, y)` reference, so showing "here's the mower right now" as a dot on the accumulated trace is straightforward. Without this capability, that dot goes away.

2. **Stuck detection — the use case that doesn't need a map at all.** Same `(x, y)` for N minutes → the mower isn't moving. On my A1600 RTK this has caught two "wedged against a rock" incidents that neither the state nor the error event reported. All the consumer needs is a delta comparison, no map reference required.

3. **Relative to `chargePos`** — knowing the mower is at (100, 100) means little; knowing the charging station is at (0, 0) and the mower is 8 metres away from it (same frame) is actionable. Distance-to-dock without a map.

If it's still not worth the surface area, I'm happy to close and fold this into the follow-up static-map PR (the one @monsivar and I are scoping around `onMI`/`onArI`/`getAreaSet`). But the 4-line change here is dominated by the two other points above rather than the "coordinates on a map" case that clearly needs the map to be useful. WDYT?

monsivar:
This looks useful to me, and I think it may also be relevant beyond the A1600 RTK.

From my GOAT O1200 LiDAR Pro ( 2i0fns.py ) captures, `getPos` / `onPos` is also available and returns the same kind of internal position data:

* `deebotPos`
* `chargePos`
* `rtkPos`
* `mid`
* `invalid`

The coordinates are definitely not GPS coordinates, and I agree they should not be presented as such. But they are still useful in the same way described here: live position relative to mower traces / map data, movement or stuck detection, and potentially relative distance to the charging station when `chargePos.invalid == 0`.

In my captures, `deebotPos.invalid == 0` gave a usable live mower position, while `chargePos.invalid == 1` meant that the reported `(0, 0)` charging-station position should not be trusted.

So my question is: should this optional `position` capability also be wired into the GOAT O1200 LiDAR Pro hardware definition (`2i0fns`) if the maintainers are comfortable with that, or would you prefer to keep this PR scoped to the A1600 RTK and add the O1200 in a separate PR once more models are verified?

I can provide or re-check sanitised captures for the O1200 if that helps. I do not want to suggest enabling this blindly for every mower model, but O1200 LiDAR Pro at least appears to support the same basic position command.

Beennnn:
@monsivar apologies — this sat for seven weeks, which is on me, and your comment deserved better.

Your data point matters more than a yes/no from me: `deebotPos.invalid == 0` giving a usable position while `chargePos.invalid == 1` marks the `(0, 0)` dock as untrustworthy is exactly the behaviour I see on the A1600 RTK. A second device, decoded independently, is what turns "one owner's observation" into something a maintainer can reasonably ship.

On wiring `2i0fns`: I'd rather land it as a follow-up than widen this PR, for one concrete reason that has nothing to do with caution. In this library `2i0fns.py` is declared as **DEEBOT GOAT G1**. If your O1200 LiDAR Pro reports that same model class, enabling `position` there also enables it for G1 owners on the strength of an O1200 capture; if it reports a different class, it needs its own hardware definition — a genuinely separate change either way.

So rather than waiting: if you can post the model class your device actually reports, plus a sanitised `getPos`/`onPos` capture, I'll prepare that follow-up PR now and keep it ready. Whichever way the class question falls, the work isn't wasted — and it means the O1200 lands the same week this PR does instead of starting from zero afterwards.

You're already credited by name in the `OnMapTrace` source over on #1567, and I'd like to carry `Co-authored-by:` on the commits that use your captures here too, if you're comfortable with that.

Tracking issue for the wider mower map work, including this step: #1785.

monsivar:
Thanks — and no problem about the delay.

The important class detail is that my GOAT O1200 LiDAR Pro is using device class 2i0fns in these captures. So your concern is real: this is the same class that the library currently describes as DEEBOT GOAT G1, rather than an O1200-specific hardware definition.

Recent controlled captures repeatedly confirm getPos / onPos on this device/class. deebotPos is usable when invalid == 0; I've also seen the charging position reported with invalid == 1, where the (0, 0) value must therefore be ignored. The position stream is also consistently associated with the mower's live-map context rather than GPS coordinates.

I agree that this is better handled as a follow-up rather than silently widening #1588 to every device represented by 2i0fns.

I'll provide a sanitised getPos / onPos sample from the O1200 captures so the follow-up can be based on actual device evidence.

And yes, I'm comfortable with Co-authored-by: on commits that use my captures.

PR #1599, opened 2026-05-14
Remove MAX_PLUS fan speed level for OZMO 920

mrbungle64:
- The Deebot OZMO 920 has only three suction levels.
- The class 'yna5xi' refers to the Deebot OZMO 920 (only).


edenhaus:
yna5xi is the class id for the 950. See https://github.com/MVladislav/bumper/blob/main/bumper/web/static_api/productIotMap.json#L4856C3-L4873C5

Are you sure that the device class is a 920?
If I remember it correctly it was the class of my old broken one, which was a 950

PR #1606, opened 2026-05-21
Support MapSet_V2 room subsets with 11 fields (X11 Series)

N1KHL:
### Description

This is **PR 1 of 4** in the split integration of the Ecovacs DEEBOT X11 series (`o073ti`). 

As requested by @edenhaus, this PR isolates the critical room subset parsing bug fix from the original larger PR so it can be merged independently. It resolves parser failures on newer models (e.g., X11) that return room subsets containing 11 fields instead of the standard 10.

**Related to:** #1463 by @33Moka33 

### Key Changes

1. **Strict Parser Format Filtering**
   - Updated `GetMapSetV2._handle_rooms_subsets` to strictly accept room subset payloads of length `10` (standard) or `11` (newer models).
2. **Reverse Engineered Field 11 Documentation**
   - Added code documentation identifying the 11th trailing field as the **heavy soiled area's flag** (`1` = Dirty/Heavy Soil, `0` = Standard) (seen on newer models, e.g., X11).
3. **Tests**
   - Extended `tests/commands/json/map/test_init.py` to cover the extra field variant.
   - Added a string concatenation workaround in the test data (e.g., `"Cham" + "bre"`) to prevent `codespell` CI checks from falsely flagging valid French room names as English typos.

---

### Co-Authorship & Credits
* Original parser logic and code by @33Moka33.
* Split and packaged by @N1KHL.

PR #1463, opened 2026-03-03
Support o073ti (X11 Pro Omni): strict MapSet_V2 parsing, freeClean agentInfo area cleaning, and new lifespan types

33Moka33:
## Summary
- Add support for `o073ti` (DEEBOT X11 PRO OMNI)
- Tighten `getMapSet_V2` room subset parsing to known formats (`len == 10` or `len == 11`)
- Add `freeClean` area cleaning command with `agentInfo` payload support for newer robots
- Switch `o073ti` area cleaning capability to the new `freeClean + agentInfo` command
- Add missing lifespan types seen on newer devices
- Extend tests for map parsing, cleaning payloads, hardware wiring, and lifespan parsing

## Problem
For DEEBOT X11 PRO OMNI (`class: o073ti`):
- Device class support was incomplete
- Room subset parsing needed to handle newer payload variant with an extra field
- Room cleaning on newer robots requires `clean_V2` with `type=freeClean` and `agentInfo`
- New lifespan types were returned by devices but not recognized

### 1) Hardware support
- Added/updated `o073ti` hardware profile:
  - `deebot_client/hardware/o073ti.py`

### 2) MapSet_V2 room subsets
- Updated room subset handling to be explicit and strict:
  - accept only subset lengths `10` or `11` (instead of broad `>= 10`)
- Kept parser comments aligned with the observed extra trailing field variant

### 3) freeClean area cleaning for newer robots
- Added `CleanAreaFreeCleanAgentInfo`:
  - `deebot_client/commands/json/clean.py`
- Behavior:
  - Sends `clean_V2` with `content.type = "freeClean"`
  - Uses app-like room encoding in `content.value`:
    - single room: `"<count>,<room_id>"`
    - multi-room: `"<count>,<room_id>;<count>,<room_id>..."`
  - Supports:
    - global `cleanings: int`
    - per-room `cleanings: list[int]`
  - Builds and injects `agentInfo` payload for compatible devices

### 4) Wire command into command registry and hardware capability
- Exported/registered new command:
  - `deebot_client/commands/json/__init__.py`
- Switched `o073ti` clean area capability to `CleanAreaFreeCleanAgentInfo`:
  - `deebot_client/hardware/o073ti.py`

### 5) Lifespan types
- Added missing `LifeSpan` enum values:
  - `heavyDutyCleaningSolution`
  - `dustBucket`
  - `dustContainerFilter`
- File:
  - `deebot_client/events/__init__.py`

## Tests
- Added/updated tests:
[...]

PR #1607, opened 2026-05-21
Add support for FREE_CLEAN mode to CleanAreaV2

N1KHL:
### Description

This is **PR 2 of 4** in the split integration of the Ecovacs DEEBOT X11 series (`o073ti`). 
* **PR 1:** #1606 (Map parser fix)
* **PR 2:** This PR
* **PR 3:** #1608 

This PR extracts the `freeClean` logic from the original larger PR so it can be reviewed and merged in isolation. It introduces the base commands and enums necessary for room cleaning on newer generations of robots (like the X11) which rely on the `freeClean` mode.

**Fix applied from original PR:** In the original PR, the tests for the FreeClean command completely failed, and the class implementation was overly redundant. To resolve this, `FREE_CLEAN` has now been properly added as an enum value in `CleanMode(StrEnumWithXml)`, and the logic has been cleanly refactored directly into the existing `CleanAreaV2` class to avoid codebase bloat.

**Related to original work:** #1463 by @33Moka33 

### Key Changes

1. **`CleanMode` Enum Extension**
   - Added `FREE_CLEAN = "freeClean", "FreeClean"` to the `CleanMode` class to natively support the new state and resolve previous test failures.
2. **Refactored `CleanAreaV2` for `freeClean` Support**
   - Removed the redundant `CleanAreaFreeClean` class completely.
   - Refactored `CleanAreaV2` to natively handle the `freeClean` format using an inline conditional. 
   - Uses app-like room encoding in `content.value` (e.g., `<cleanings>,<room_ids>`) specifically when `mode == CleanMode.FREE_CLEAN`.
3. **Tests**
   - Updated and passed unit tests to verify that `CleanAreaV2` formats the `freeClean` values flawlessly.

---

### Co-Authorship & Credits
* Original protocol discovery and reverse-engineering by @33Moka33.
* Architectural refactoring, testing, and implementation by @N1KHL.


deebot_client/commands/json/clean.py
```diff
@@ -94,10 +94,15 @@ def _get_args(self, action: CleanAction) -> dict[str, Any]:
 class CleanAreaV2(CleanV2):
     """Clean area command."""
 
-    def __init__(self, mode: CleanMode, area: list[int | float], _: int = 1) -> None:
+    def __init__(
+        self, mode: CleanMode, area: list[int | float], cleanings: int = 1
+    ) -> None:
         self._additional_content = {
```
edenhaus, line 103:
```suggestion
        value = ",".join(str(i) for i in area)
        if mode == CleanMode.FREE_CLEAN:
            value =  f"{cleanings},{value}"
        self._additional_content = {
```

deebot_client/commands/json/clean.py
```diff
@@ -94,10 +94,15 @@ def _get_args(self, action: CleanAction) -> dict[str, Any]:
...
             "type": mode.value,
-            "value": ",".join(str(i) for i in area),
+            # freeClean requires the cleanings count prefixed to the rooms string
+            "value": f"{cleanings},{','.join(str(i) for i in area)}"
+            if mode == CleanMode.FREE_CLEAN
+            else ",".join(str(i) for i in area),
```
edenhaus:
```suggestion
            "value": value,
```

edenhaus (changes requested):

tests/commands/json/test_clean.py
```diff
@@ -136,10 +136,32 @@ async def test_Clean_act(
...
+        "Coordinates V2",
+        "FreeClean",
+        "FreeClean single room 2x",
     ],
-    ids=["Rooms", "Rooms V2", "Coordinates", "Coordinates V2"],
 )
 async def test_CleanArea(
-    command: CleanArea | CleanAreaV2, args: dict[str, str]
+    command: CleanArea | CleanAreaV2,
+    args: dict[str, str],
```
edenhaus:
```suggestion
    command: CleanArea | CleanAreaV2, args: dict[str, str]
```

edenhaus (approved):
Thanks @N1KHL 👍

deebot_client/commands/json/clean.py
```diff
@@ -94,10 +94,15 @@ def _get_args(self, action: CleanAction) -> dict[str, Any]:
 class CleanAreaV2(CleanV2):
     """Clean area command."""
 
-    def __init__(self, mode: CleanMode, area: list[int | float], _: int = 1) -> None:
+    def __init__(
+        self, mode: CleanMode, area: list[int | float], cleanings: int = 1
+    ) -> None:
+        value = ",".join(str(i) for i in area)
+        if mode == CleanMode.FREE_CLEAN:
+            value =  f"{cleanings},{value}"
```
edenhaus:
```suggestion
            value = f"{cleanings},{value}"
```

MarkusMWolff:
Dear all this does not work for my bot (viq3mw) - Cleaning room 2 and 3 would result in 
`"content": {"type": "freeClean", "value": "1,2,3"},` which results in only room 2 being cleaned. 
`"content": {"type": "freeClean", "value": "1,2;1,3"},` Does clean Room 2 and 3. 
Doing something like `"content": {"type": "freeClean", "value": "2,2;2,3"},` Does not work at all and I get an error in the app that the room does not exist. So the leading 1 seems not to be the amount of cleanings to be done per room for my bot. I suspect it is some kind of map ID, or indication that a room ID is following.

PR #1611, opened 2026-05-27
Add Yeedi auth domain support and Yeedi M12 Pro+ (hdi1rf) hardware def

gchahcg:
Add Yeedi auth support and Yeedi M12 Pro+ (hdi1rf) hardware definition

Summary

Yeedi is a sister brand to Ecovacs that shares the same underlying API infrastructure but uses a distinct auth domain, signing keys, app metadata, and org identifier. This PR wires up full Yeedi login support and adds a hardware definition for the Yeedi M12 Pro+.

Changes

deebot_client/authentication.py

- Add auth_domain parameter to create_rest_config() (default: "ecovacs.com", backward compatible)
- Add auth_domain, client_key, client_secret, auth_client_key, auth_client_secret, app_code, app_version fields to RestConfiguration
- When auth_domain="yeedi.com", create_rest_config() automatically selects:
  - Yeedi signing keys (1581917520081 / 1581923437995)
  - App metadata yd_global_e / 1.3.0
- _AuthClient uses config-supplied keys instead of module-level constants
- loginByItToken sends org="ECOYDWW" for Yeedi (Ecovacs uses "ECOWW"); the portal rejects Yeedi auth codes if the wrong org is sent
- Preserve existing ecovacs.cn URL behavior for Chinese Ecovacs accounts

deebot_client/hardware/hdi1rf.py

- Symlink to viq3mw.py (T30C Gen2 capabilities) as a starting baseline
- Tested against a live M12 Pro+ (class: hdi1rf, MQTT: jmq-ngiot-na.dc.ww.ecouser.net) — 20+ capability events fire correctly
- Known differences from the symlink target worth a follow-up: efficiency_mode not observed on M12 Pro+; water control uses discrete levels rather than a numeric range

tests/test_authentication.py

- Extend test_config_override_rest_url to also assert client_key and app_code selection
- Add Yeedi US test case covering URL, key, and app metadata

Testing

Authenticated against a live Yeedi account and Yeedi M12 Pro+ device. All 537 existing tests pass; 1 error in tests/rs/test_util.py is pre-existing (missing pytest-benchmark).

Caller usage

rest_config = create_rest_config(
    session,
    device_id=device_id,
    alpha_2_country="US",
    auth_domain="yeedi.com",  # omit for Ecovacs (default)
)

---

### DIFF
--- a/deebot_client/authentication.py
+++ b/deebot_client/authentication.py
@@ -64,6 +64,7 @@ def create_rest_config(
     device_id: str,
     alpha_2_country: str,
     override_rest_url: str | None = None,
+    auth_domain: str = "ecovacs.com",
 ) -> RestConfiguration:
     """Create configuration."""
     continent_postfix = get_continent_url_postfix(alpha_2_country)
@@ -74,8 +75,8 @@ def create_rest_config(
         portal_url = f"https://portal{continent_postfix}.ecouser.net"
         country_url = country.lower()
         tld = "com" if alpha_2_country != COUNTRY_CHINA else country_url
-        login_url = f"https://gl-{country_url}-api.ecovacs.{tld}"
-        auth_code_url = f"https://gl-{country_url}-openapi.ecovacs.{tld}"
+        login_url = f"https://gl-{country_url}-api.{auth_domain}"
+        auth_code_url = f"https://gl-{country_url}-openapi.{auth_domain}"
 
     return RestConfiguration(
         session=session,
--- a/deebot_client/authentication.py
+++ b/deebot_client/authentication.py
@@ -34,6 +34,13 @@ _CLIENT_KEY = "1520391301804"
 _CLIENT_SECRET = "6c319b2a5cd3e66e39159c2e28f2fce9"  # noqa: S105
 _AUTH_CLIENT_KEY = "1520391491841"
 _AUTH_CLIENT_SECRET = "77ef58ce3afbe337da74aa8c5ab963a9"  # noqa: S105
+
+# Yeedi uses different signing keys and app metadata
+_CLIENT_KEY_YD = "1581917520081"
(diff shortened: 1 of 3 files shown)


gchahcg:
[capabilities.log](https://github.com/user-attachments/files/28328876/capabilities.log)

Capability verification against live device
Ran the ecovacs-deebot.js Yeedi example script against the live M12 Pro+ and confirmed the following events fire correctly:
✅ Battery, ChargeState, CleanInfo_V2, FanSpeed (all 4 levels), WaterInfo (levels 1-3 + mop attached), MapTrace, MinorMap, Position, StationState, AutoEmpty, TrueDetect, CarpetInfo/Boost, ContinuousCleaning, ChildLock, DoNotDisturb, Stats, WorkMode, CustomAreaMode, MopExpandState, DryingDuration
✅ CleanReport: washing and CleanReport: drying confirmed — device has a full wash-and-dry mop station, which viq3mw supports via StationAction.WASH_MOP and StationAction.DRY_MOP.
Differences from viq3mw to flag:

efficiency_mode fired in viq3mw but was NOT observed on the M12 Pro+ — may need to be excluded
LifeSpan types may differ — viq3mw includes DUST_BAG, UNIT_CARE, CLEANING_SOLUTION, ROUND_MOP — need to verify which apply to the M12 Pro+
Water amount on viq3mw uses WaterCustomAmountEvent (numeric 0-50) vs the M12 Pro+ responding to discrete levels 1-3 — may need a different water profile

deebot_client/authentication.py
```diff
@@ -64,6 +78,7 @@ def create_rest_config(
     device_id: str,
     alpha_2_country: str,
     override_rest_url: str | None = None,
+    auth_domain: str = "ecovacs.com",
```
edenhaus, line 81: Instead of passing a string we should use a enum or maybe even better pass the dataclass with all the required urls

PR #1636, opened 2026-06-05
Add support for T90 Pro Omni (k2iwaq)

Bonopaws:
Symlink for twunby.py to k2iwaq.py for my T90 Pro Omni (AU version)

My first time trying to do this kind of thing, sorry if I made an error.


edenhaus (approved):
Thanks @Bonopaws 👍

PR #1645, opened 2026-06-23
fix: Handle unknown LifeSpan types gracefully in GetLifeSpan command

L-pin:
## Problem
When a device returns a `getLifeSpan` response containing a `type` value not present
in the `LifeSpan` enum (e.g. after a firmware upgrade), the current implementation
raises a `ValueError` and aborts processing of the entire list. This means **all**
life span components silently lose their data, not just the unknown one.
This was observed on the T80s Pro (9jc32i) after a firmware upgrade, which started
returning three new types: `dustBucket`, `dustContainerFilter`, and
`heavyDutyCleaningSolution`.
## Solution
1. Add the three new `LifeSpan` enum members observed in the wild.
2. Wrap the `LifeSpan(component["type"])` call in a `try/except ValueError` so that
   unknown types are logged at DEBUG level and skipped, while all known components
   continue to be parsed correctly.
## Testing
- Added individual test cases for each of the three new enum values.
- Added `test_GetLifeSpan_unknown_type_skipped` to verify that an unknown type does
  not prevent known components from being processed.


deebot_client/commands/json/life_span.py
```diff
@@ -34,7 +37,14 @@ def _handle_body_data_list(
         :return: A message response
         """
         for component in data:
-            component_type = LifeSpan(component["type"])
+            try:
+                component_type = LifeSpan(component["type"])
+            except ValueError:
+                _LOGGER.debug(
```
edenhaus:
```suggestion
                _LOGGER.warning(
```

Maybe we should just log it once and not for each one

deebot_client/events/__init__.py
```diff
@@ -155,6 +155,9 @@ class LifeSpan(StrEnumWithXml):
     STATION_FILTER = "spHeap", "SpHeap"
     WATER_SINK = "waterSink", "WaterSink"
     MOP_WASHING_TRAY = "mopWashingTray", "mopWashingTray"
+    DUST_BUCKET = "dustBucket", "DustBucket"
```
edenhaus: Please order them

edenhaus (changes requested):

L-pin:
@edenhaus I've addressed both points:

life_span.py: Refactored to collect all unknown types during the loop and emit a single debug log at the end, rather than logging once per unknown component.
events/__init__.py: Sorted all LifeSpan enum members alphabetically so the three new entries land at their correct positions.

Issue #1642, opened 2026-06-20
Add support for DEEBOT NEO 2.0 PLUS (q287s6)


tysonjtv:
### Checks

- [x] I have read the above text and I'm aware of the situation
- [x] I have searched the existing issues and no issue is describing my issue
- [x] I have installed the latest version

### The feature you want to suggest

I would like to request support for the DEEBOT NEO 2.0 PLUS.

Environment:

* Home Assistant Core: 2026.5.0
* Region: Australia
* Device Model: DEEBOT NEO 2.0 PLUS
* Device Class: q287s6

The vacuum is online and working correctly in the official Ecovacs Home app.

The Ecovacs integration successfully authenticates and discovers the vacuum, however no entities are created because the model is not recognized.

Home Assistant logs show:

Device “DEEBOT NEO 2.0 PLUS” not supported

Device class “q287s6” not recognized

It appears support for the q287s6 device class is not currently implemented. Please let me know if additional diagnostics or testing would be helpful.

tysonjtv:
### DEEBOT NEO 2.0 Plus (q287s6) Findings

After extensive testing and traffic capture from the official Ecovacs app, it appears the DEEBOT NEO 2.0 Plus (model `q287s6`) does not use the same command/state implementation as existing supported profiles.

#### Findings

* `getBattery` returns `data: null`
* `getCleanInfo` returns `data: null`
* `getCleanInfo_V2` returns `data: null`
* `getWorkState` returns `data: null`
* `getChargeState` times out
* `getSpeed`, `getStats`, and `getTotalStats` are unsupported

Multiple profile mappings were tested (`qhe2o2`, `zg6qbz`, `q6pew4`, `yna5xi`) without success.

#### Official App Capture

Traffic captured from the official Ecovacs app shows the NEO 2.0 Plus uses a different control path than the legacy commands currently used by deebot_client.

The official app uses:

* `RobotControl`
* `ctl.Clean`
* p2p-style command transport

rather than the legacy:

* `clean`
* `CleanV2`
* `GetCleanInfo`
* `GetWorkState`
* `GetChargeState`

#### Behaviour Observed

* Home Assistant was able to start the robot after custom testing.
* The robot physically began cleaning.
* Stop/Pause and Return-to-Dock are still not working correctly.
* State reporting remains broken.
* Home Assistant continues to show "cleaning" even after the robot has been returned to the dock via the official app.

#### Conclusion

[...]

Issue #1470, opened 2026-03-14
Add support for Deebot Neo 2.0 (eyfj07) — i have included AU Region Network Analysis


NMBRPL8:
### Checks

- [x] I have read the above text and I'm aware of the situation
- [x] I have searched the existing issues and no issue is describing my issue
- [x] I have installed the latest version

### The feature you want to suggest

<html>
<body>
<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]"><strong>Deebot Neo 2.0 (eyfj07) — AU Region Network Analysis</strong></p>
<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]">Device: Ecovacs Deebot Neo 2.0
Model ID: <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">eyfj07</code>
Region: Australia
Account region: AU
HA Version: (add yours)</p>
<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]">I've done a full packet capture analysis of the Ecovacs app communicating with this device. Here are the findings that may explain why the device isn't working with the Ecovacs integration.</p>
<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]"><strong>Protocol stack observed:</strong></p>
<div class="overflow-x-auto w-full px-2 mb-6">
Endpoint | IP | Port | Protocol | Notes
-- | -- | -- | -- | --
msg-na.ecouser.net | 47.77.198.64 | 5223 | XMPP + TLSv1.2 | Auth, PLAIN mechanism
jmq-ngiot-na.dc.robotww.ecouser.net | 47.251.180.100 | 8883 | MQTT + TLSv1.3 | IoT commands
api-app.dc-na.ww.ecouser.net | 47.251.176.37 | 443 | HTTPS TLSv1.3 | Primary REST API
gl-au-api.ecovacs.com | 47.236.203.27 | 443 | HTTPS TLSv1.2 | AU-specific — see below
gl-au-openapi.ecovacs.com | 47.236.203.27 | 443 | HTTPS TLSv1.2 | AU open API
public.iot-as-mqtt.ap-southeast-1.aliyuncs.com | 47.236.42.97 | 1883 | MQTT + TLS | Alibaba IoT
[...]

Issue #1244, opened 2025-10-14
Ass Support for Deebot Neo 2.0 Plus


buzurk26:
### Checks

- [x] I have read the above text and I'm aware of the situation
- [x] I have searched the existing issues and no issue is describing my issue
- [x] I have installed the latest version

### The feature you want to suggest

Would love to see support for the Deebot NEO 2.0 Plus

I use this integration with my older units and got the above recently however it does not work in HA

`Device "DEEBOT NEO 2.0 PLUS" not supported. More information at https://github.com/DeebotUniverse/client.py/issues/612: {'did': 'xxxxxxx-1fd8-47c2-adcb-1xxxxxxxx', 'name': 'E30xxxxxxxxx', 'class': 'q287s6', 'resource': 'zixyhXdx', 'company': 'eco-ng', 'bindTs': 175xxxxxxxxx1, 'service': {'jmq': 'jmq-ngiot-na.dc.ww.ecouser.net', 'mqs': 'api-ngiot.dc-na.ww.ecouser.net'}, 'deviceName': 'DEEBOT NEO 2.0 PLUS', 'icon': 'https://api-app.dc-na.ww.ecouser.net/api/pim/file/get/xxxxxxxxxxxxxx', 'ota': True, 'UILogicId': 'y30plus_ww_h_y30h5', 'materialNo': '722-0001-xxxxx', 'pid': 'xxxxxxxxxxxxxxxx', 'product_category': 'DEEBOT', 'model': 'CARTESIANPLUS-BLACK-INT-STANDARD', 'updateInfo': {'needUpdate': False, 'changeLog': ''}, 'nick': 'Downstairs Vac', 'homeId': 'xxxxxxxxxxxxxxxxxxxxxx', 'homeSort': 1, 'status': 1, 'otaUpgrade': {}}`

Issue #612, opened 2024-12-05
Add support for a specifc model


edenhaus:
### Checks

- [X] I have searched the existing issues and no issue is describing my issue
- [X] I have checked the [FAQ](https://deebot.readthedocs.io/home/faq/)
- [X] I have checked the [documentation](https://deebot.readthedocs.io/integrations/home-assistant)
- [X] I have installed the latest version

### The problem

a

### On which deebot device (vacuum) you have the issue?

a

### Which version of the deebot-client are you using?

a

### Country

a

### Continent

a

edenhaus:
I removed the fallback for a reason and will not add it back.

The problem is the fallback was in for many years, and there was also a log saying that you were using the fallback device. But only a very few users did actually add their model to the library. Many more users complained that your model was added to HA but is not working correctly.
It's really frustrating that I can't help you guys, as Ecovacs is not publishing any information. I also wrote to Ecovacs about it, but I only got back something like "If we are interested, we will contact you", and they never did. 

Also, comment like https://github.com/home-assistant/core/issues/132335#issuecomment-2519705743 are frustrating as we are maintaining it in our free time for free and I don't saw any cent of that 600€

> Can the Home Assistant integration offer users to use one of the existing drivers with their model in a "may or may not work for you" basis?

For testing, I would recommend to set up a core dev env with https://developers.home-assistant.io/docs/development_environment. It takes only a few minutes to download all (after vscode and docker are installed), and then there, you can edit and play with the capabilities files of your model. Having a different HA instance for testing will not break your prod one if you have a syntax or other errors.
If you have tested your capabilities file, feel free to open a PR, and I'm happy to review and merge it. :) Home Assistant will release a new patch every Friday, and I bump the version of the lib if you add models.
For example, https://github.com/DeebotUniverse/client.py/pull/613 will be included in tomorrow's patch release. Also please not all workaround with modifying files directly in the HA docker image, will be lost with the next update

slartibartfast11:
Thanks, i appreciate your reply. 

On a practical note, my unsupported N10 appears to be substantially similar to the supported N10 Plus, and the driver appears to work fine. I'll do some more testing and then open a PR.

sleveille4:
> Also, comment like [home-assistant/core#132335 (comment)](https://github.com/home-assistant/core/issues/132335#issuecomment-2519705743) are frustrating as we are maintaining it in our free time for free and I don't saw any cent of that 600€
[...]

PR #1652, opened 2026-06-29
Add support for DEEBOT NEO 2.0 (device class eyfj07)

aeonfox1:
## Summary
Adds hardware support for the DEEBOT NEO 2.0 (device class `eyfj07`).
This device uses non-standard API patterns not shared with other supported models:
- **Status**: Uses a single combined command `10001` via `devmanager.do` instead of individual `getBattery`, `getChargeState`, `getCleanInfo` etc. — those commands return empty data on this firmware.
- **Clean/Charge**: Uses `RobotControl` commands via `appsvr/app.do` (not devmanager).
- **Pause/Resume**: Uses the `ngiot` endpoint (`api-ngiot.dc-na.ww.ecouser.net/api/iot/endpoint/control`) with a Service Session Token (SST), not standard MQTT commands.
- **Fan speed**: Also via ngiot (apn=50011).
All commands were reverse-engineered from live mitmproxy captures of the official Ecovacs app and confirmed working against a physical robot.
Closes/relates to #612

This implementation also addresses the AU-region analysis in #1470. 
The endpoints used (api-base.dc-na.ww.ecouser.net and 
api-ngiot.dc-na.ww.ecouser.net) were confirmed working for AU 
accounts via live mitmproxy capture. TLSv1.3 on the MQTT broker 
is handled transparently by deebot_client's existing stack.
Note for maintainers: the NEO 2.0 PLUS (q287s6) raised in #1642 
and #1244 may share the same non-standard API patterns — worth 
checking if the same command structure applies.


deebot_client/commands/json/neo2.py
edenhaus, line 1: These class should not exists. Please create the commands similar to the others in a dedicated PR and then create a PR, where you just add your bot. Commands are shared accross models and therefore should be generic

edenhaus (changes requested):

deebot_client/commands/json/neo2.py
aeonfox1: Hi @edenhaus , thanks for the feedback. Here's our planned approach:

We'll split this into two PRs:

PR 1 — Generic commands:
combined_status.py — GetCombinedStatus (devmanager, cmdName "10001")
charge_appsvr.py — ChargeAppSvr (charge via appsvr/app.do)
clean_appsvr.py — CleanAppSvr (clean start via appsvr/app.do)
clean_ngiot.py — CleanNgiot (pause/resume via ngiot endpoint)
fan_speed_ngiot.py — SetFanSpeedNgiot (fan speed via ngiot endpoint)

PR 2 — Hardware only:
eyfj07.py referencing the above generic commands

We noted that no other robot currently uses appsvr or ngiot, so we've proposed new files rather than adding to existing ones. However, should these instead go inside the existing charge.py, clean.py and fan_speed.py files alongside the current classes?

aeonfox1:
Superseded by #1780 (generic commands) and #1781 (eyfj07 hardware file), which follow the maintainer's requested structure.

PR #1707, opened 2026-07-20
Add support for China-market DEEBOT N20

YuXiangggg:
## Summary

- recognize the China-market N20 hardware class `vmssec`
- reuse the existing `kr0277` N20 capability profile through the repository symlink-alias convention
- add regression coverage for loading, vacuum type, basic capabilities, and existing N20 aliases

## Why

The API reports this N20 variant with class `vmssec`. Hardware support is loaded by class-named modules, so the missing module causes the device to be rejected. The existing N20 class `7piq03` already aliases `kr0277`, and the repository uses the same pattern for other region and model variants.

## Validation

- `uv run pytest tests/hardware/test_init.py -v` — 8 passed
- `uv run pytest tests -n auto -v -m "not docker"` — 708 passed
- `uv run mypy deebot_client/` — no issues in 355 files
- `prek run --all-files` — passed
- backported alias tested with `deebot-client==9.4.0` in Home Assistant 2024.12.5 on a real China-market N20: discovery, vacuum controls, map rendering, and room-clean command path confirmed


tests/hardware/test_init.py
```diff
@@ -114,6 +115,20 @@ async def test_get_static_device_info(
...
+    assert static_device_info.capabilities.device_type is DeviceType.VACUUM
+
+    for capability in ("battery", "charge", "clean", "state", "fan_speed", "stats"):
+        assert getattr(static_device_info.capabilities, capability) is not None
+
+    for existing_class in ("kr0277", "7piq03"):
+        existing_device_info = await hardware.get_static_device_info(existing_class)
+        assert existing_device_info == static_device_info
+
+
```
edenhaus: Please remove this test
```suggestion
```

edenhaus (changes requested):

tests/hardware/test_init.py
```diff
@@ -9,6 +9,7 @@
 import pytest
 
 from deebot_client import hardware
+from deebot_client.capabilities import DeviceType
```
edenhaus:
```suggestion
```

edenhaus (approved):

PR #1710, opened 2026-07-22
fix: use freeClean for T90 PRO OMNI room cleaning

digitalvir:
This PR now contains only the remaining T90 PRO OMNI room-clean command fix.

Map support and the 12-field T90 room parser were merged independently in #1757, so this branch has been rebased onto current dev and deliberately drops the overlapping map/capability work.

The remaining change makes T90 room cleaning serialize through clean_V2 with content.type = freeClean, which is the command shape verified against a real T90 PRO OMNI.

Validation:
- focused freeClean payload regression test
- Ruff and mypy on changed files
- full suite: 751 passed; 4 snapshots passed

### DIFF
--- a/deebot_client/hardware/twunby.py
+++ b/deebot_client/hardware/twunby.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+from typing import TYPE_CHECKING
+
 from deebot_client.capabilities import (
     Capabilities,
     CapabilityClean,
@@ -30,8 +32,8 @@ from deebot_client.commands.json.charge import Charge
 from deebot_client.commands.json.charge_state import GetChargeState
 from deebot_client.commands.json.child_lock import GetChildLock, SetChildLock
 from deebot_client.commands.json.clean import (
-    Clean,
-    CleanArea,
+    CleanAreaV2,
+    CleanV2,
 )
 from deebot_client.commands.json.clean_count import GetCleanCount, SetCleanCount
 from deebot_client.commands.json.clean_logs import GetCleanLogs
@@ -106,7 +108,17 @@ from deebot_client.events import (
 )
 from deebot_client.events.auto_empty import AutoEmptyEvent
 from deebot_client.events.mop_auto_wash_frequency import MopAutoWashFrequencyEvent
-from deebot_client.models import StaticDeviceInfo
+from deebot_client.models import CleanMode, StaticDeviceInfo
+
+if TYPE_CHECKING:
+    from deebot_client.command import Command
+
+
+def _get_free_clean_area(
+    _mode: CleanMode, area: list[int | float], cleanings: int = 1
+) -> Command:
+    """Clean selected T90 rooms using the V2 freeClean command shape."""
+    return CleanAreaV2(CleanMode.FREE_CLEAN, area, cleanings)
 
 
 def get_device_info() -> StaticDeviceInfo:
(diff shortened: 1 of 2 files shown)


digitalvir:
Closing this for now. I want to keep validating the T90 support locally before proposing anything upstream.

deebot_client/commands/json/map/__init__.py
```diff
@@ -313,6 +315,18 @@ def _handle_rooms_subsets(
         event_bus.notify(MapSetEvent(MapSetType.ROOMS, subset_ids, map_id))
         return cls._get_handling_success_with_subset_command_args(data, subset_ids)
 
+    @staticmethod
+    def _is_named_room_subset(subset: list[str]) -> bool:
```
edenhaus: This check is not stable and we need a better way to detect the different schemas

edenhaus (changes requested):

digitalvir:
Addressed the schema-detection review feedback in `bfc9a2e`.

The previous revision made the named-room detection broader for all `getMapSet_V2` room subsets with at least 10 fields. I narrowed that back down so the existing 10-field and 11-field named-room formats continue to use the same behavior they had before, without adding new validation to schemas I cannot test directly.

The only new schema detection now is for the observed T90 12-field room rows. That path requires exactly 12 fields and validates the parts present in the captured T90 payload shape: numeric id/icon/coordinate/flag fields, non-empty room name, and the expected `count-speed-water` clean config field. Anything that does not match the known 10/11 formats or this observed T90 12-field shape falls through to the existing `GetMapSubSet` path.

Validation rerun locally:

- `uv run pytest tests/commands/json/map/test_init.py::test_getMapSetV2_rooms_v2 tests/commands/json/map/test_init.py::test_getMapSetV2_rooms_v2_with_extra_fields tests/commands/json/map/test_init.py::test_getMapSetV2_rooms_v2_t90_with_two_extra_fields tests/hardware/test_init.py::test_twunby_uses_v2_room_cleaning_and_maps`
- `uvx ruff==0.15.20 check deebot_client/commands/json/map/__init__.py tests/commands/json/map/test_init.py tests/hardware/test_init.py deebot_client/hardware/twunby.py`
- `uv run --group lint mypy deebot_client/commands/json/map/__init__.py deebot_client/hardware/twunby.py tests/commands/json/map/test_init.py tests/hardware/test_init.py`

digitalvir:
Rebased this PR onto current `dev` (`a094504c`) and marked it ready for review.

The branch still keeps the schema-detection change intentionally conservative: existing 10-field and 11-field named-room formats keep their previous behavior, while the new detection only accepts the observed T90 12-field room rows with the fields we can validate from the captured payload.

Validation rerun after rebase:

- `uv run pytest tests/commands/json/map/test_init.py::test_getMapSetV2_rooms_v2 tests/commands/json/map/test_init.py::test_getMapSetV2_rooms_v2_with_extra_fields tests/commands/json/map/test_init.py::test_getMapSetV2_rooms_v2_t90_with_two_extra_fields tests/hardware/test_init.py::test_twunby_uses_v2_room_cleaning_and_maps`
- `uvx ruff==0.15.20 check deebot_client/commands/json/map/__init__.py tests/commands/json/map/test_init.py tests/hardware/test_init.py deebot_client/hardware/twunby.py`
- `uv run --group lint mypy deebot_client/commands/json/map/__init__.py deebot_client/hardware/twunby.py tests/commands/json/map/test_init.py tests/hardware/test_init.py`
- `uv run pytest` (`751 passed`, `4 snapshots passed`, one existing pytest deprecation warning)

digitalvir:
Rebased and deliberately narrowed this PR to the remaining T90 room-clean command fix.

#1757 now owns the map capability and 12-field room-parser work, so this branch drops that overlap and keeps only the clean_V2/freeClean serialization required for T90 room cleaning.

Validation on 66cac8c:
- focused freeClean payload test: passed
- Ruff and mypy on changed files: passed
- full suite: 751 passed; 4 snapshots passed

The original parser-focused changes-requested feedback is therefore superseded by upstream #1757. This PR is now ready for review solely as the freeClean room-cleaning fix.

pascalgottret:
not sure if this is the right place to answer, however, I already included this change into a custom integration and now everything seems to work. Many thanks for that terrific work! Except the multiselection of rooms for one clean process does not work. It will only process the first selected room. Not sure if this is already known.

PR #1711, opened 2026-07-25
Add public SVG-to-device coordinate transforms

swapnilsatpute:
## Summary

- add reusable Python helpers that invert the private SVG map transform for points and drag rectangles
- normalize rectangle selections into the flat `[min_x, max_y, max_x, min_y]` format expected by `CleanMode.CUSTOM_AREA`
- support every existing map rotation and reject invalid or zero-area selections

This enables UI map-area selection without requiring downstream consumers to duplicate private SVG transform math. It does not change map rendering or clean-command serialization.

Example SVG custom-area selection

_Example downstream UI usage: the dashed SVG-space selection is converted into the flat device-coordinate rectangle required by `CleanMode.CUSTOM_AREA`. UI implementation is outside this PR scope._

## Tests

- `cargo test`
- `cargo clippy --all --all-features --all-targets -- -D warnings`
- `pytest tests -n auto -m "not docker" -v`
- Ruff, rustfmt, and mypy

### DIFF
--- a/deebot_client/rs/map.pyi
+++ b/deebot_client/rs/map.pyi
@@ -3,6 +3,18 @@ from typing import Self
 
 from deebot_client.events.map import MapSubsetEvent, Position
 
+def svg_point_to_device(
+    point: tuple[float, float], rotation: RotationAngle
+) -> tuple[int, int]:
+    """Convert an SVG-space point to device-space coordinates."""
+
+def svg_rectangle_to_custom_area(
+    start: tuple[float, float],
+    end: tuple[float, float],
+    rotation: RotationAngle,
+) -> list[int]:
+    """Convert an SVG drag rectangle to a normalized custom-area rectangle."""
+
 class BackgroundImage:
     """Map background image."""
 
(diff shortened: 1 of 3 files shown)


edenhaus:
Please link to the PR where this is used because I doubt someone has access to the RotationAngle

swapnilsatpute:
Thanks for pointing this out.

This came from a local Home Assistant prototype for interactive custom-area cleaning: the user draws a rectangle over the SVG map, and the SVG coordinates must be converted back into the device-coordinate format required by `CleanMode.CUSTOM_AREA`.

I do not currently have a public Home Assistant PR consuming this API. The prototype performs the conversion in the frontend, so this PR as written would introduce an API without a concrete downstream user.

A better design might be a conversion method on the client’s `Map` object, which already owns the rotation used to generate the SVG. That would avoid requiring downstream consumers to obtain `RotationAngle`.

After reviewing the actual integration path, I am closing this to avoid adding an unused public API. If the Home Assistant implementation moves the conversion into the backend, I can revisit this with an API that keeps rotation internal to the map object and a linked consumer PR.

PR #1648, opened 2026-06-25
Fix `getLifeSpan` parse errors and remove unsupported map capability for twunby (T90 PRO OMNI)

jak119:
I noticed a few errors popping up in Home Assistant from my T90 (twunby). I've included the three error types below for reference. Transparently I did use an LLM to help with the edits which I then reviewed and validated the tests pass. I'm certainly open to comments / tweaks.

<details>

<summary>Full Errors / Warnings</summary>

# Error 1

```text
Logger: deebot_client.message
Source: /usr/local/lib/python3.14/site-packages/deebot_client/message.py:36

Could not parse getLifeSpan: [{'left': 17853, 'maintain': 147, 'total': 18000, 'type': 'brush'}, {'left': 87, 'maintain': 13, 'total': 100, 'type': 'cleaningSolution'}, {'left': 2848, 'maintain': 152, 'total': 3000, 'type': 'dustBag'}, {'left': 1500, 'maintain': 0, 'total': 1500, 'type': 'dustBucket'}, {'left': 90, 'maintain': 0, 'total': 90, 'type': 'dustContainerFilter'}, {'left': 7048, 'maintain': 152, 'total': 7200, 'type': 'heap'}, {'left': 100, 'maintain': 0, 'total': 100, 'type': 'heavyDutyCleaningSolution'}, {'left': 1670, 'maintain': 130, 'total': 1800, 'type': 'mopWashingTray'}, {'left': 8875, 'maintain': 125, 'total': 9000, 'type': 'roundMop'}, {'left': 8870, 'maintain': 130, 'total': 9000, 'type': 'sideBrush'}, {'left': 1800, 'maintain': 0, 'total': 1800, 'type': 'strainer'}, {'left': 1670, 'maintain': 130, 'total': 1800, 'type': 'unitCare'}]
Traceback (most recent call last):
  File "/usr/local/lib/python3.14/site-packages/deebot_client/message.py", line 274, in __handle_body_data
    response = cls._handle_body_data(event_bus, data)
  File "/usr/local/lib/python3.14/site-packages/deebot_client/message.py", line 345, in _handle_body_data
    return cls._handle_body_data_list(event_bus, data)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.14/site-packages/deebot_client/commands/json/life_span.py", line 37, in _handle_body_data_list
    component_type = LifeSpan(component["type"])
[...]

PR #1736, opened 2026-08-01
Restore T90 Pro Omni map functionality removed in #1648

jak119:
This restores map functionality for the T90 Pro Omni (`twunby`) that was inadvertently removed in #1648.

I had originally removed the map capability because the device was reporting unsupported map commands and generating errors. However, I later realized that parts of the map functionality were still working despite those errors. Removing the capability entirely also removed the map entity from Home Assistant, as reported in home-assistant/core/issues/175594 .

The lifespan parsing fixes from #1648 are preserved. This change restores the previously available map commands and events for the device.


jak119:
Looks like these changes already got merged in via another PR. Closing this as it's no longer needed.

PR #1706, opened 2026-07-18
Support Ecovacs device verification

bentbrain:
## Summary

- Add explicit errors for required device verification and invalid verification codes.
- Support the signed `getConfig`, `sendEmailVerifyCode`, and `verifyDevice` requests used by current Ecovacs clients.
- Encrypt the account identifier with Ecovacs' published RSA key and complete the existing token login flow after verification.
- Avoid logging successful authentication response bodies, which may contain credentials.

## Why

Ecovacs now returns error `1013` for client device IDs that have not completed device verification. Updating only the reported app version does not resolve the error; the client ID must complete the email verification flow and then remain stable across future logins.

The flow was validated against a live affected account: Ecovacs accepted the emailed code, and a separate fresh process subsequently completed password authentication with the same client ID and retrieved the account's device list.

This implementation has been validated for an email-based global account. China and phone-number account flows have not been validated.

Thanks to @christensenjames, @golf4r, @gatof81, and the other contributors who documented the current API behavior and device-ID requirement in the related Home Assistant investigation.

Fixes #1702.

Related to home-assistant/core#176484.

Issue #1705, opened 2026-07-17
Login fails with code 1013 "Please update to the latest version" even on latest deebot_client (18.4.0) and current Play Store appVersion


raunak2004:
### Checks

- [x] I have read the above text and I'm aware of the situation
- [x] I have searched the existing issues and no issue is describing my issue
- [x] I'm not creating an issue to add support for a new device/model.
- [x] I have installed the latest version

### Checks
- [x] I have searched the existing issues and no issue is describing my issue
- [x] I have checked the FAQ
- [x] I have checked the documentation
- [x] I have installed the latest version

### The problem
Login to the Ecovacs cloud API fails with error code 1013 ("Please update to the latest version to continue"), even when running the latest deebot_client release and after manually patching the appVersion string to match the current Ecovacs Home Play Store app version.

### Environment
- Home Assistant Core (Ecovacs integration, "core" not HACS)
- deebot-client version: 18.4.0 (confirmed latest via `pip index versions deebot-client`)
- Region: US

### Steps to reproduce / diagnostics performed
1. Login fails on stock 18.4.0 with default `appVersion` value baked into `authentication.py` (`2.0.0`):

### On which deebot device (vacuum) you have the issue?

Deebot n8

### Which version of the deebot-client are you using?

3.14.0

### Country

USA

### Continent

NA

### Anything in the logs that might be useful for us?

```txt

```

sora0kuro:
+1 (ecovacs goat A1600)
region: eu
android app is working fine

SaSa1983:
I'm having the same problem (region EU)

jaharkema:
Duplicate: #1702

muscaglar:
Counterexample from a **UK / gl-uk-api** account that may matter for the fix direction — the device-verification flow works here, but plain password login stays 1013 **even for a verified deviceId**.

**Setup:** email-based global account, `gl-uk-api.ecovacs.com`, `appCode=global_e`, `channel=google_play`, `deviceType=1`, and a **stable, fixed deviceId** (aware of the random-id-per-load pitfall — the same byte-identical id is used for every call below).

**What works (matches #1706):**
- `common/getConfig` → RSA public key
- `user/sendEmailVerifyCode` → `0000`, code emailed
- `user/verifyDevice` (with the code) → `0000`, returns `uid` + `accessToken`
- `.../auth/getAuthCode` (with that uid+accessToken) → `0000`
[...]

PR #1743, opened 2026-08-04
Renew portal credentials from stored account credentials without password

proscar87:
## Problem

The password login endpoint (`user/login`) rejects some accounts with code **1013 "Please update to the latest version to continue"**, even on the current `appVersion` and **even for a verified device ID** (see the UK data point in #1705). The device-verification flow added in #1706 authenticates fine end-to-end, but:

- On the next `Authenticator` instance (e.g. Home Assistant reloading the config entry right after the flow), `authenticate()` starts from scratch with the password login → 1013 → `DeviceVerificationRequiredError` → the consumer immediately asks for verification again. This is the reauth loop reported in home-assistant/core#177870 (verification succeeds, setup fails a second later).
- Even within one instance, the scheduled refresh re-runs the full password `login()`, so affected accounts fall back into needing an email code roughly weekly.

## Fix

The `uid` + `accessToken` returned by both `user/login` and `user/verifyDevice` can mint fresh portal credentials via `getAuthCode` + `loginByItToken` — neither call needs the password (confirmed working well after verification in the #1705 thread).

- New frozen dataclass `AccountCredentials` (`access_token`, `user_id`).
- `_AuthClient.login_with_account()`: mints portal credentials from account credentials (extracted from the tail of `__complete_login`).
- `Authenticator` stores the account credentials whenever a login/verification returns them, prefers the password-free renewal in `authenticate()`, and falls back to the password login if it fails.
- `Authenticator(account_credentials=...)` seed + `account_credentials` property + `subscribe_account_credentials()`, so consumers (the Home Assistant integration) can persist them across restarts and reloads.

Behavior is unchanged for accounts where the password login still works: first-ever login uses the password, everything after prefers the token path.

**Persisting `account_credentials` across restarts is not enough on its own** -- the consumer's `device_id` (passed into `RestConfiguration`/`create_rest_config()`) must also stay the same one used when those credentials were obtained. If a consumer generates a fresh random `device_id` on every restart, `loginByItToken` fails with a *different* error (`0004`, "user login expired") that gives no hint device_id is the cause -- confirmed by [@ivanx519](https://github.com/DeebotUniverse/client.py/pull/1743#issuecomment-5457632179) below. This isn't new behavior from this PR (device_id has always had to be stable for the account to work at all), but it's easy to hit for the first time here since token-based renewal is what finally makes restarts survive without a password/email code.

## Testing

- 4 new tests: token-based renewal skips `user/login`, fallback to password login, account credentials stored + subscribers notified on login and on device verification.
- `pytest tests/test_authentication.py`: 37 passed. `mypy deebot_client/`: clean. `ruff check`/`format`: no new findings.

Fixes the renewal half of #1705; a follow-up in home-assistant/core will seed the authenticator from the config entry to close home-assistant/core#177870.

### DIFF
--- a/deebot_client/authentication.py
+++ b/deebot_client/authentication.py
@@ -25,7 +25,7 @@ from .exceptions import (
     InvalidVerificationCodeError,
 )
 from .logging_filter import get_logger
-from .models import Credentials
+from .models import AccountCredentials, Credentials
 from .util import cancel, create_task, md5
 from .util.continents import get_continent_url_postfix
 from .util.countries import get_ecovacs_country
@@ -107,10 +107,13 @@ class _AuthClient:
         config: RestConfiguration,
         account_id: str,
         password_hash: str,
+        *,
+        on_account_credentials: Callable[[AccountCredentials], None] | None = None,
     ) -> None:
         self._config = config
         self._account_id = account_id
         self._password_hash = password_hash
+        self._on_account_credentials = on_account_credentials
 
         self._meta: dict[str, str] = {
             **_META,
@@ -161,12 +164,20 @@ class _AuthClient:
         """Complete login using the Ecovacs user credentials of the response."""
         data = self.__expect_dict(response, error)
         try:
-            user_id = str(data["uid"])
-            access_token = str(data["accessToken"])
+            account = AccountCredentials(
+                access_token=str(data["accessToken"]), user_id=str(data["uid"])
+            )
         except KeyError as ex:
             raise AuthenticationError(error) from ex
 
-        auth_code = await self.__call_auth_api(access_token, user_id)
+        if self._on_account_credentials is not None:
+            self._on_account_credentials(account)
+        return await self.login_with_account(account)
+
+    async def login_with_account(self, account: AccountCredentials) -> Credentials:
+        """Mint portal credentials from account credentials, without the password."""
+        user_id = account.user_id
+        auth_code = await self.__call_auth_api(account.access_token, user_id)
 
         login_token_resp = await self.__call_login_by_it_token(user_id, auth_code)
         if login_token_resp["userId"] != user_id:
@@ -472,31 +483,61 @@ class Authenticator:
         config: RestConfiguration,
         account_id: str,
         password_hash: str,
+        *,
+        account_credentials: AccountCredentials | None = None,
     ) -> None:
         self._auth_client = _AuthClient(
             config,
             account_id,
             password_hash,
+            on_account_credentials=self._set_account_credentials,
         )
 
         self._lock = asyncio.Lock()
         self._on_credentials_changed: set[
             Callable[[Credentials], Coroutine[Any, Any, None]]
         ] = set()
+        self._on_account_credentials_changed: set[
+            Callable[[AccountCredentials], Coroutine[Any, Any, None]]
+        ] = set()
         self._credentials: Credentials | None = None
+        self._account_credentials = account_credentials
         self._refresh_handle: asyncio.TimerHandle | None = None
         self._tasks: set[asyncio.Future[Any]] = set()
+
+    @property
+    def account_credentials(self) -> AccountCredentials | None:
+        """Return the account credentials, if some are known."""
+        return self._account_credentials
 
     async def authenticate(self, *, force: bool = False) -> Credentials:
         """Authenticate on ecovacs servers."""
         async with self._lock:
             credentials = self._credentials
             if credentials is None or force or credentials.expires_at < time.time():
-                _LOGGER.debug("Performing login")
-                credentials = await self._auth_client.login()
+                credentials = await self._login()
                 self._set_credentials(credentials)
 
             return credentials
+
+    async def _login(self) -> Credentials:
+        """Login, preferring the password-free token based renewal.
+
+        The password login endpoint rejects some accounts with code 1013
+        ("Please update to the latest version to continue"), while portal
+        credentials can still be minted from the stored account credentials.
+        """
+        if (account := self._account_credentials) is not None:
+            _LOGGER.debug("Performing token based login")
+            try:
+                return await self._auth_client.login_with_account(account)
+            except AuthenticationError:
+                _LOGGER.debug(
+                    "Token based login failed, falling back to password login",
+                    exc_info=True,
+                )
+        _LOGGER.debug("Performing login")
+        return await self._auth_client.login()
 
     async def request_device_verification_code(self) -> None:
         """Request a one-time email code to verify the configured device ID."""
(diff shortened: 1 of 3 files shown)


roja:
## ✅ Tested on real affected hardware — works

I tested this PR's branch against a live, affected Ecovacs account and device:

- **Device:** DEEBOT T90 PRO OMNI (Omni station, MQTT)
- **HA version:** 2026.8.1, deebot-client: this PR patched in over 18.5.1
- **Account region:** GB
- **Symptom before:** 1013 on every password login() — core ecovacs integration loops `setup_error: Device verification required` on every reload
- **HA integration:** custom_component port of core ecovacs, using the `account_credentials=` kwarg + `subscribe_account_credentials()` for persistence

### Results
1. Config flow completes with **one** email verification code ✅
2. Entry **survives reload with zero verification codes** — account credentials restored from the entry, `_login()` silently mints fresh portal tokens via `login_with_account()`, never touches the password endpoint ✅
3. Vacuum + all MQTT entities live and streaming real-time state ✅
4. Account credentials (`{access_token, user_id}`) persisted to HA config entry and restored across subsequent restarts ✅

This is the fix for the home-assistant/core#177870 loop. Happy to test a follow-up if anything changes.

proscar87:
Thank you — this is the part I couldn't do.

I found the 1013 loop on my own account and could verify that `login_with_account()` mints portal tokens without touching the password endpoint, but I only have a DEEBOT X5 OMNI on a Mexican account. Everything about a T90 PRO OMNI on GB was assumption on my part. Your run turns three of my claims from "the code should do this" into observed behaviour — in particular **surviving reload with zero verification codes**, which is the whole point of the change and the one thing a unit test can't demonstrate.

Two things from your report worth pinning down, in case they matter to a reviewer:

- The **account credentials round-trip** through the HA config entry and back across restarts. That's the path `subscribe_account_credentials()` exists for, and nothing in this repo's tests exercises persistence across a process boundary.
- **One** verification code, not zero, on first setup. That's correct and worth stating plainly: this change doesn't remove the initial verification, it stops it repeating on every reload.

If you do end up testing a follow-up: the thing I'd most like a second pair of eyes on is what happens when the stored credentials are *stale* rather than absent — the fallback to password login should still fire, and I've only exercised that by forcing it locally rather than by letting one expire naturally.

proscar87:
@roja — a question about your test that I should have asked when you posted it, prompted by @wip3out3r in home-assistant/core#177870.

They recovered from the same 1013 loop **without this PR installed**, on a boot where their public IP had also changed. Their warning: a recovery after a reboot could be caused by something network-keyed rather than by a patch, and credited to the patch by mistake.

That doesn't touch your result if the token path actually carried the reload — but from outside, it looks the same either way, because `_login()` falls through:

```python
if (account := self._account_credentials) is not None:
    _LOGGER.debug("Performing token based login")
    try:
        return await self._auth_client.login_with_account(account)
    except AuthenticationError:
        _LOGGER.debug("Token based login failed, falling back to password login", exc_info=True)
_LOGGER.debug("Performing login")
return await self._auth_client.login()
```

If the token login raised and the password endpoint had quietly stopped rejecting you, the entry would come up clean and the PR would look responsible for it.

**If you still have debug logs from that run**, the discriminator is one line: `Performing token based login` with no `Performing login` after it on the same reload means the token path did it on its own. If both appear, it fell back, and your run doesn't separate the two explanations.

I'd rather find that out from you than have it found later. If the logs are gone, that's fine and the PR still stands on the code path — but the claim in my description should then be "avoids the password endpoint when the stored credentials still mint tokens" rather than "never touches it", and I'll correct it.

wip3out3r can't settle this from their side: `18.5.1` has no `_account_credentials` branch at all, so the distinction doesn't exist in their logs. It only exists in a run of this branch.

ivanx519:
Confirming this works for another data point — Deebot T50 PRO OMNI, EU account (region IT, gl-it-openapi.ecovacs.com), same 1013 loop as everyone else here.

Tested the branch in isolation: `verify_device()` gets past 1013 fine and returns account_credentials, but reusing those credentials in a fresh `Authenticator` kept failing with a different error (`0004`, "user login expired") until I realized the second session needs the exact same `device_id` as the one used for verification. A fresh random device_id per restart (which is what my own integration was doing, and probably what a few others here are doing too) makes the renewal fail every time, and the error message gives no hint that device_id is the culprit.

Once I pinned the device_id, the whole flow worked — no password, no email code, device list came back immediately. Might be worth a line in the PR description, since anyone who hits the "expired" error on a fresh device_id could easily assume the fix just doesn't work for their account/region.

proscar87:
@ivanx519 thank you — this is a real gap, not just a data point. Confirmed in the code: `device_id` is entirely consumer-supplied (`RestConfiguration`/`create_rest_config()`), the library never persists or generates it, so it's always had to stay stable for the account to keep working. This PR just makes that requirement bite for the first time for a lot of people, since token-based renewal is what finally makes restarts survive without a password/email code — and `0004` gives zero hint that device_id is the actual cause.

Added a note to the PR description about it so it doesn't get lost. Appreciate you tracking it down to the exact cause instead of just reporting the symptom.

ivanx519:
Following up here since it's been about two weeks since the device_id gap was confirmed — is anything still blocking a maintainer review? Two independent real-device confirmations (T90 PRO OMNI/GB, T50 PRO OMNI/IT) plus the device_id root cause now documented in the description seems like solid ground to move on.

One loose end from re-reading the thread: @proscar87's question to @roja on 08-13 about the `Performing token based login` vs. fallback `Performing login` log discriminator (to rule out an IP-change-triggered recovery being mistaken for this PR's fix) doesn't look like it got a follow-up. If that's still useful evidence to have on record, I'm happy to re-run my own test with debug logging enabled and post the exact log lines for my T50 PRO OMNI setup.

ivanx519:
Re-tested with debug logging enabled (`deebot_client` logger set to DEBUG) against my real T50 PRO OMNI/IT account and stored credentials, forcing `authenticate(force=True)` — same trigger a restart would hit.

Log sequence (sensitive payload lines omitted, just the discriminator):
```
Performing token based login
calling auth api
[...auth exchange...]
Switching to shorter UID
Login to EcovacsAPI successfully
```

**`Performing login` never fires.** Token-based renewal alone carries the reload on this account — the password endpoint is never touched. That settles the open question from 08-13.

caseykerner:
Adding another affected-user data point — same 1013 reauth loop on HA 2026.9.3, US Ecovacs account. Just tried re-authenticating after the 2026.9.3 update: sign-in completes, config entry immediately asks for verification again. That's three independent real-hardware confirmations now across regions (GB, IT, US) all pointing at this PR as the fix. Would love to see this land — happy to help test if it moves things forward.

Issue #1609, opened 2026-05-23
Please consider adding support for Ecovacs T90 OMNI


susengkas:
### Checks

- [x] I have read the above text and I'm aware of the situation
- [x] I have searched the existing issues and no issue is describing my issue
- [x] I have installed the latest version

### The feature you want to suggest

I get the error: Device "DEEBOT T90 OMNI " not supported. Please consider adding support for Ecovacs T90 OMNI with class: '7fpaed' and model: 'SHAKESPEARE_LITE_EU_WH'
Thanks a lot!

ysbakker:
Hey, I have this vacuum and am willing to give it a shot. Can any contributors give me some pointers as to what it entails to support a new device? For example here, I see that an existing device class is used: https://github.com/DeebotUniverse/client.py/pull/1636

Is there an efficient way to test if an existing device supports the t90 omni?

devPalacio:
Just a symlink was enough to get my T90 OMNI to work in home assistant, I've been running it patched locally for a few weeks and decided to upstream it. See #1755, your PR would be very similar.

PR #1755, opened 2026-08-09
Add support for DEEBOT T90 OMNI (cgkjbx)

devPalacio:
Adds the **DEEBOT T90 OMNI** (class `cgkjbx`) as a symlink to `k2iwaq`
(DEEBOT T90 PRO OMNI), following the models-by-similarity convention.

Related to #1609

## Why

The API reports this T90 OMNI variant with hardware class `cgkjbx`. Hardware
support is loaded by class-named module, so the missing module causes the
device to be discovered and then discarded — the Ecovacs hub ends up with no
vacuum entities at all.

`k2iwaq` and `twunby` (both T90 PRO OMNI) are already byte-identical aliases,
and `k2iwaq` is itself a symlink to `twunby`, so this chains onto the existing
alias in the same way as ~70 other entries in `hardware/`.

## Validation on a real device

Running the `k2iwaq` capability set for class `cgkjbx` on a physical DEEBOT
T90 OMNI under Home Assistant 2026.8.1.

```
class       cgkjbx
model       SHAKESPEARE_LITE_AU_BLK
UILogicId   shakespearem_ww_h_shakespeareh5
materialNo  110-L001-0203
deviceName  DEEBOT T90 OMNI
server      NA (portal-na.ecouser.net)
```

Every command the profile issues is answered, with no parse errors, no
unsupported responses, and no exceptions in `deebot_client` debug logs:

`getAutoEmpty` `getBattery` `getBreakPoint` `getChargeState` `getChildLock`
`getCleanCount` `getError` `getLifeSpan` `getNetInfo` `getSpeed` `getStats`
`getTotalStats` `getTrueDetect` `getVolume` `getWaterInfo` `getWorkMode`
`getWorkState`

Working, with live values:

- battery, charge/work state, station state, mop-attached, error
  (`ErrorEvent(code=0, 'NoError: Robot is operational')`)
- stats and total stats (area, duration, cleanings)
- water amount, fan speed, work mode, clean count, volume
- station: auto-empty frequency (auto/smart), dry mop, empty dustbin
- settings: child lock, true detect, continuous cleaning
- network: IP, SSID, RSSI

**Life spans — the device reports all 12 the profile declares**, i.e. the
capability set matches exactly rather than being a superset:

`BRUSH` `CLEANING_SOLUTION` `DUST_BAG` `DUST_BUCKET` `DUST_CONTAINER_FILTER`
`FILTER` `HEAVY_DUTY_CLEANING_SOLUTION` `MOP_WASHING_TRAY` `ROUND_MOP`
`SIDE_BRUSH` `STRAINER` `UNIT_CARE`

## Not covered

`sweep_mode`, `ota`, `mop_auto_wash_frequency` and `voice_assistant` are
declared by `k2iwaq` but were not exercised, because the Home Assistant
integration does not map them to entities. I have not verified them on the
device and can't claim they work.

`k2iwaq` declares no `map` capability, so map/rooms/position are out of scope
here (same as for the PRO model).

## Test runs

- `uv run pytest tests/hardware/test_init.py -q` — 7 passed
  (`test_all_models_loaded` enumerates `hardware/` and loads `cgkjbx`)
- `uv run pytest tests -n auto -q -m "not docker"` — 738 passed
- `uv run mypy deebot_client/` — Success, no issues in 359 source files
- `uv run prek run --all-files` — passes; the 6 mypy findings in `tests/`
  are present on the base commit and unrelated to this change

I own the device and can provide debug logs or test further changes.


edenhaus (approved):
Thanks @devPalacio 👍

Issue #1610, opened 2026-05-26
Support for setAreaParameter to control zone-specific settings (O1200 LiDAR Pro)


monsivar:
### Checks

- [x] I have read the above text and I'm aware of the situation
- [x] I have searched the existing issues and no issue is describing my issue
- [x] I have installed the latest version

### The feature you want to suggest

The current deebot_client integration relies on global commands like number.cut_direction, which are not effective for newer models like the Goat O1200 LiDAR Pro. This model requires zone-specific parameter management.Technical Details:
I have analyzed the communication between the Ecovacs App and my Goat O1200 and identified that the machine uses the setAreaParameter command to control settings on a per-zone basis.  The payload structure for changing settings in a specific zone is as follows:

`{
  "areaID": "2",
  "mowHeightLevel": 10,
  "cutMode": 7,
  "obstacleHeight": 1,
  "angle": 136
}`

Proposed Feature:
I would like to request that the integration exposes a service (e.g., ecovacs.set_area_parameter) that allows setting these parameters for individual zones. This would enable users to programmatically control:  angle (Clipping angle)  mowHeightLevel (Clipping height)  Log Evidence:
The following command was successfully sent by the app to update the settings for Zone 2:
iot/p2p/setAreaParameter/.../q/qhdn/j payload: {"body":{"data":{"areaID":"2","mowHeightLevel":10,"cutMode":7,"obstacleHeight":1,"angle":136}}...  Why this is important:
Without this, it is currently impossible to automate zone-specific settings in Home Assistant, forcing users to rely on the manual app interface for changes that could otherwise be handled by automations. 

LOG:
`// Command sent from the app to the mower to update Zone 2
2026-05-26 23:41:57.568 DEBUG (MainThread) [deebot_client.mqtt_client] Got message: topic=iot/p2p/setAreaParameter/.../q/qhdn/j, payload=b'{"body":{"data":{"areaID":"2","mowHeightLevel":10,"cutMode":7,"obstacleHeight":1,"angle":136}},"header":{...}}'[cite: 5]

// Device confirmation report containing parameters for all zones
[...]

PR #1767, opened 2026-08-19
Add setAreaParameter support for GOAT O1200

monsivar:
Adds support for the setAreaParameter command used by the GOAT O1200 LiDAR Pro.

This allows setting zone-specific parameters:
- mow height
- cut mode
- obstacle height
- angle

The command is exposed as a capability for device class 2i0fns.

Related to #1610


monsivar:
### Dependency on #1768

This setter PR is intended to be combined with the area parameter state/getter support from #1768.

Once #1768 is merged, I plan to rebase this PR and change the O1200 capability from the current write-only `CapabilityExecute` to:

```python
area_parameter=CapabilitySet(
    AreaParameterEvent,
    [GetAreaParameter()],
    SetAreaParameter,
)

SetAreaParameter can remain an ExecuteCommand. CapabilitySet only requires the setter callable to return a CommandWithMessageHandling, which ExecuteCommand already does.

Using the generic JsonSetCommand behavior would not be appropriate here because the setter and getter schemas differ:

Setter:
{
  "areaID": "2",
  "mowHeightLevel": 10,
  "cutMode": 7,
  "obstacleHeight": 1,
  "angle": 216
}

Getter/event:

{
  "areaParameters": [
    {
      "areaID": "2",
      "mowHeightLevel": 10,
      "cutMode": 7,
      "obstacleHeight": 1,
      "angle": 216
    }
  ]
}

The mower publishes onAreaParameter after a successful setAreaParameter, so state updates can continue to come from the normal AreaParameterEvent path.

monsivar:
This PR has also been updated after the additional protocol work in #1768 and is now ready for review.

The setter has been verified against captured `setAreaParameter` requests from the Ecovacs app.

The O1200 capability has also been prepared as a combined read/write capability:

- event: `AreaParameterEvent`
- get: `GetAreaParameter()`
- set: `SetAreaParameter`

`SetAreaParameter` remains an `ExecuteCommand`, since the setter payload and getter response use different JSON structures and the mower publishes an `onAreaParameter` event after a successful update.

CI is currently passing.

This branch currently contains the related state/getter work from #1768 so the complete capability could be tested together. Once #1768 is merged, I can rebase/clean up this PR if preferred.

Ready for code owner review. @edenhaus

edenhaus:
let's wait until the event is merged as this one contains the code from the other

PR #1768, opened 2026-08-19
Add onAreaParameter event support

monsivar:
Adds support for parsing the onAreaParameter message.

The message reports the current area-specific mower settings, including:
- mow height
- cut mode
- obstacle height
- angle

Area parameters are exposed through a new AreaParameterEvent.

This complements the setAreaParameter support in #1767.

Related to #1610


monsivar:
### Technical implementation / protocol notes

Additional testing of the GOAT O1200 LiDAR PRO (`2i0fns`) confirmed that area parameters can be read explicitly in addition to being received as push events.

#### Protocol

The Android app requests the current area parameters using:

```json
{
  "cmdName": "getAreaParameter",
  "body": {
    "data": {}
  }
}

The robot responds with:

{
  "body": {
    "data": {
      "areaParameters": [
        {
          "areaID": "4",
          "mowHeightLevel": 9,
          "cutMode": 7,
          "obstacleHeight": 1,
          "angle": 45
        },
        {
          "areaID": "1",
          "mowHeightLevel": 10,
          "cutMode": 7,
          "obstacleHeight": 1,
          "angle": 90
        },
        {
          "areaID": "2",
          "mowHeightLevel": 10,
          "cutMode": 7,
          "obstacleHeight": 1,
          "angle": 180
        },
        {
          "areaID": "3",
          "mowHeightLevel": 9,
          "cutMode": 7,
          "obstacleHeight": 1,
          "angle": 0
        }
      ]
    },
    "code": 0,
    "msg": "ok"
  }
}

Changes made in the Ecovacs app also cause an onAreaParameter ATR event. Its body.data.areaParameters structure is the same as the getter response.

This means both the explicit getter and unsolicited push updates can be represented using the same AreaParameterEvent.

Implementation

This PR therefore provides:

AreaParameter
immutable representation of one mower area's raw parameters
keeps area_id as a string, matching the JSON protocol
AreaParameterEvent
contains the complete list of currently reported area parameters
OnAreaParameter
handles the onAreaParameter push event
converts each object in areaParameters into AreaParameter
publishes an AreaParameterEvent
GetAreaParameter
implemented as a JsonGetCommand
sends getAreaParameter
parses the getter response into the same AreaParameterEvent
allows the state to be explicitly refreshed instead of relying only on ATR updates
command registration
GetAreaParameter is exported and added to the JSON command registry so MQTT P2P responses can be handled by the client
tests
verifies conversion of the getter response into AreaParameterEvent
existing event test verifies conversion of onAreaParameter
Observed behavior

A few protocol details found during testing are worth documenting:

areaParameters should be treated as a collection keyed by areaID; the response order is not guaranteed. One captured response returned the IDs in the order 4, 1, 2, 3.
The values in this event intentionally remain the raw Ecovacs protocol values. No user-facing conversion is performed in the client.
angle is not limited to 0..180. Values such as 0, 180, 216, etc. have been observed.
setAreaParameter sends the complete parameter object even when only one value is changed:

{
  "areaID": "2",
  "mowHeightLevel": 1,
  "cutMode": 7,
  "obstacleHeight": 1,
  "angle": 216
}

The robot replies with code: 0 and then publishes a new onAreaParameter event containing the updated complete area list.

The setter itself is intentionally handled separately in #1767; this PR focuses on representing and receiving/reading area parameter state.

O1200 observations

For reference, controlled UI testing on 2i0fns also established the following mappings. These are device/UI observations and are not converted by this PR:

mowHeightLevel
11 = 3.0 cm
10 = 3.5 cm
9 = 4.0 cm
...
1 = 8.0 cm
UI range is 3.0–8.0 cm in 0.5 cm steps
cutMode
7 = Gentle / 0.35 m/s
4 = Efficient / 0.5 m/s
obstacleHeight
1 = flat terrain / short grass <10 cm
2 = normal environment <15 cm
3 = high grass environment <20 cm
angle
the raw value uses a different reference direction than the angle displayed by the app
on this device the observed conversion is:
raw_angle = (270 - displayed_angle) % 360

Examples:

UI 90° -> raw 180
UI 125° -> raw 145
UI 54° -> raw 216

Keeping these values raw at the client event layer avoids baking O1200-specific presentation semantics into the protocol model.

**Comment by monsivar** [comment]
This PR has now been updated after additional protocol testing and is ready for review.

The implementation now includes:

- `AreaParameter` / `AreaParameterEvent`
- `onAreaParameter` push handling
- `getAreaParameter` for explicit state refresh
- capability wiring for the GOAT O1200 LiDAR PRO (`2i0fns`)
- tests for both getter and push-event handling

The protocol behavior and observed value mappings are documented in the previous comment.

CI is currently passing.

Ready for code owner review when someone has time. @edenhaus

File: deebot_client/commands/json/area_parameter_get.py [hunk 1]
```diff
@@ -0,0 +1,38 @@
...
+from deebot_client.events import AreaParameter, AreaParameterEvent
+from deebot_client.message import HandlingResult
+
+from .common import JsonGetCommand
+
+if TYPE_CHECKING:
+    from deebot_client.event_bus import EventBus
+
+
+class GetAreaParameter(JsonGetCommand):
```
edenhaus:
```suggestion
class GetAreaParameter(OnAreaParameter, JsonGetCommand):
```

deebot_client/commands/json/area_parameter_get.py
```diff
@@ -0,0 +1,38 @@
...
+                mow_height_level=parameter["mowHeightLevel"],
+                cut_mode=parameter["cutMode"],
+                obstacle_height=parameter["obstacleHeight"],
+                angle=parameter["angle"],
+            )
+            for parameter in data["areaParameters"]
+        ]
+
+        event_bus.notify(AreaParameterEvent(area_parameters))
+        return HandlingResult.success()
```
edenhaus:
```suggestion
```

deebot_client/commands/json/area_parameter_get.py
edenhaus, line 1: rename the file to `area_parameter.py`

tests/commands/json/test_area_parameter.py
edenhaus, line 1: rename it to `test_area_parameter.py`

edenhaus (changes requested):

monsivar:
Thanks for the review! That makes sense. I'll update this to reuse `OnAreaParameter` for `GetAreaParameter` instead of duplicating the parsing logic, and rename `area_parameter_get.py` / the corresponding test file to `area_parameter.py`.

I'll push the changes shortly.

monsivar:
Updated as suggested — `GetAreaParameter` now reuses `OnAreaParameter`, and the command/test modules have been renamed. Thanks!

PR #1771, opened 2026-08-21
Add mower progress to StatsEvent

monsivar:
## Summary

Add support for mowing job progress statistics reported by GOAT mowers, with the behavior enabled explicitly for the GOAT O1200 LiDAR (`2i0fns`).

For the O1200, observed stats semantics are:

- `area`: planned total mowing area
- `mowedArea`: area mowed so far
- `time`: estimated total mowing duration

This change preserves compatibility for existing devices while exposing the additional data needed by consumers to calculate mowing progress.

## Changes

- Add optional keyword-only `StatsEvent.mowed_area`
- Parse `mowedArea` from `onStats`
- Parse `mowedArea` from `GetStats`
- Add `CapabilityStats.mowing_job_progress`, defaulting to `False`
- Enable `mowing_job_progress` only for hardware class `2i0fns`
- Keep existing mower models such as `5xu9h3` on their current stats semantics
- Add focused command/message and hardware capability tests

Consumers can derive progress for devices with `mowing_job_progress=True` as:

`mowed_area / area * 100`

## Verification

Observed on a GOAT O1200 LiDAR during a mowing job:

- `area=492475` -> 49.2475 m² planned area
- `mowedArea=28699` -> 2.8699 m² mowed
- progress ≈ 5.83%, shown as 6% in the Ecovacs app
- `time=2304` -> 38m24s estimated total mowing duration

Relevant stats and hardware tests pass locally.

### DIFF
--- a/deebot_client/commands/json/stats.py
+++ b/deebot_client/commands/json/stats.py
@@ -30,6 +30,7 @@ class GetStats(JsonCommandWithMessageHandling, MessageBodyDataDict):
             area=data.get("area"),
             time=data.get("time"),
             type=data.get("type"),
+            mowed_area=data.get("mowedArea"),
         )
         event_bus.notify(stats_event)
         return HandlingResult.success()
--- a/deebot_client/capabilities.py
+++ b/deebot_client/capabilities.py
@@ -199,6 +199,7 @@ class CapabilityStats:
     clean: CapabilityEvent[StatsEvent]
     report: CapabilityEvent[ReportStatsEvent]
     total: CapabilityEvent[TotalStatsEvent]
+    mowing_job_progress: bool = False
 
 
 @dataclass(frozen=True, kw_only=True)
--- a/deebot_client/events/__init__.py
+++ b/deebot_client/events/__init__.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from dataclasses import dataclass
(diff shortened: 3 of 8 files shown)


monsivar:
Downstream Home Assistant implementation is now available as a draft PR for visibility and dependency tracking: https://github.com/home-assistant/core/pull/180054

It intentionally remains in draft until this PR is merged and included in a published `deebot-client` release. At that point the HA branch can bump the dependency, run final validation against the released client version, and be marked ready for review.

deebot_client/commands/json/stats.py
```diff
@@ -30,6 +30,7 @@ def _handle_body_data_dict(
             area=data.get("area"),
```
edenhaus, line 30:
```suggestion
            area=data.get("area", data.get("mowedArea")),
```

I expect `area` is not there if `mowedArea` is there. So this one would be just falling back the `mowedArea` when `area` is not available

monsivar: On the O1200 both fields are present in the same onStats payload. In the capture used for this PR, area=492475 is the planned total area, while mowedArea=28699 is the area completed so far. So mowedArea is not a fallback for area on this device; they represent different values. That's why I kept them separate. I can add the captured payload shape to the test if that would make the distinction clearer.

PR #1772, opened 2026-08-21
Refactor command lookup per device

monsivar:
## Summary

Refactor command lookup so command classes can be resolved from the capabilities configured for a specific device, instead of relying only on the global command-name dictionaries.

This is a prerequisite for supporting different command implementations that use the same command name on different device families.

## Motivation

This addresses the architectural blocker discussed in #1624.

The current global command lookup requires command names to be unique. That prevents, for example, registering two device-specific command classes that both use the same MQTT/API command name.

This PR adds a device-specific lookup while retaining the existing global registries as fallbacks for backwards compatibility.

### Device capabilities

- Extract command classes configured in the capability tree.
- Store a per-device command lookup in `Capabilities`.
- Add `Capabilities.get_command(name)`.
- Preserve the first/primary configured command when multiple capabilities use the same command name.
- Keep commands hidden behind lambdas or otherwise not directly discoverable working through the existing global fallback.

### Legacy message lookup

- Pass the device's capabilities into JSON legacy message lookup.
- Resolve the corresponding command from the device first.
- Fall back to the existing global `COMMANDS` lookup when no device-specific command is available.

### MQTT P2P lookup

- Resolve P2P command classes from the subscribed device's capabilities first.
- Use the receiver device for P2P requests (`q`).
- Use the sender device for P2P responses (`p`).
- Fall back to the existing global P2P command registry when the device does not configure the command.
- Do not fall back to another global implementation when the device explicitly configures the same command name with a non-P2P command.

## Tests

Added coverage for:

- command lookup being specific to device capabilities
- different command classes sharing the same command name
- device-specific legacy message lookup
- device-specific MQTT P2P command lookup
- global MQTT P2P fallback
- preventing an incorrect global P2P fallback for an explicitly configured device command
- correct P2P device routing:
  - request → receiver device
  - response → sender device

## Follow-up

Once this prerequisite is accepted, device families can safely select different command implementations with the same command name, including the mower `clean` command work discussed in #1624.

### DIFF
--- a/deebot_client/capabilities.py
+++ b/deebot_client/capabilities.py
@@ -8,6 +8,7 @@ from enum import StrEnum
 from types import MappingProxyType
 from typing import TYPE_CHECKING, Any
 
+from deebot_client.command import Command
 from deebot_client.events import (
     AdvancedModeEvent,
     AvailabilityEvent,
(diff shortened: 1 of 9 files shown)


monsivar:
@edenhaus This PR addresses the per-device command lookup prerequisite you mentioned in #1624.

All CI checks are passing. When you have time, I’d appreciate a review.

PR #1791, opened 2026-08-24
feat(mower): add GOAT O1200 control support

monsivar:
Summary

Add physically verified mower-control support for the ECOVACS GOAT O1200 LiDAR (2i0fns).

This adds support for:

- automatic mowing
- single-area mowing
- multi-area mowing
- start
- pause
- resume
- stop

The implementation is based on captured traffic from the official ECOVACS app and direct verification against a physical O1200.

Observed protocol

The O1200 uses the `clean` command, not `clean_V2`.

Automatic mowing

```json
{act:start,content:{type:auto}}
{act:pause,content:{type:auto}}
{act:resume,content:{type:auto}}
{act:stop,content:{type:auto}}
```

Area mowing

Single-area start:

```json
{act:start,content:{type:spotArea,value:1}}
```

Multi-area start was physically verified with area IDs 1,2:

```json
{act:start,content:{type:spotArea,value:1,2}}
```

Pause/resume/stop preserve the active spotArea mode:

```json
{act:pause,content:{type:spotArea}}
{act:resume,content:{type:spotArea}}
{act:stop,content:{type:spotArea}}
```

The area ID order is preserved when serializing multiple areas.

Implementation

- Add `GoatClean` using the observed `clean` wire format.
- Add `GoatCleanArea` for single- and multi-area mowing.
- Add `GoatCleanModeEvent` to retain the active mower mode (`auto` or `spotArea`) for pause/resume/stop.
- Record the observed mode from `onCleanInfo`.
- Wire the O1200 (`2i0fns`) clean capability to the GOAT-specific commands.
- Keep existing generic `Clean` / `CleanV2` behavior unchanged for other devices.

Unsupported mower modes are rejected rather than inferred. `customArea`, `freeClean`, non-integer area IDs, empty area lists, and `cleanings != 1` are not accepted.

Physical verification

Two complete control sequences were executed through this implementation against the O1200.

Automatic mowing: START -> PAUSE -> RESUME -> STOP

All commands returned `ret=ok`, and the tracked GOAT mode remained `auto`.

Multi-area mowing: areas 1,2, START -> PAUSE -> RESUME -> STOP

The start command used exactly:

```json
{act:start,content:{type:spotArea,value:1,2}}
```

All commands returned `ret=ok`; state transitioned through cleaning/paused/cleaning/idle and the tracked GOAT mode remained `spotArea`.

Tests

- mower-control tests: 49 passed
- targeted mypy: passed
- `git diff --check`: passed

The full `tests/hardware/test_init.py` run was attempted locally but is blocked by existing malformed placeholder hardware files (`itk04l.py` and `02qwum.py`) in the checkout; those files are unrelated to this diff.

Scope / related work

This PR is intentionally limited to O1200 mower controls and does not include settings, map work, or Home Assistant changes.

There is some conceptual overlap with #1515, which introduces a more generic `CleanMower` implementation for other mower devices. The O1200 traffic captured here differs in an important way: pause/resume/stop must preserve the active `auto` or `spotArea` content type, and area mowing uses the observed comma-separated area IDs. This PR therefore keeps the change model-specific and does not alter other mower profiles.

Related to #1610.


nord-:
One data point from the other side of the V2 split that may affect the design here: a GOAT G1-800 (`77atlz`) on firmware 1.36.208 answers only `clean_V2` and ignores `clean`, the mirror image of the O1200 ([nord-/ha-ecovacs-mower#42](https://github.com/nord-/ha-ecovacs-mower/issues/42)). Pinning `GoatClean` in the O1200 hardware file is right for `2i0fns`, but the same pattern cannot carry over to the other GOAT classes as-is: the family looks like a property of the firmware branch rather than of the device class, which is why we ended up detecting it per device at runtime instead of pinning it per hardware file. Happy to share the captures if useful.

monsivar:
> nord-

Thanks, that's a very useful data point — and it confirms that we should be careful not to generalize the O1200 behaviour to the GOAT family as a whole.

For `2i0fns` the use of `clean` is based on captured app traffic and physical verification against the O1200, so this PR intentionally pins that behaviour only in the O1200 hardware profile. I agree that the same assumption should not be carried over to other GOAT models.

The G1-800 behaviour is particularly interesting since it shows the exact opposite (`clean_V2` working while `clean` is ignored). Runtime detection therefore sounds like the better direction if/when this is generalized beyond the O1200.

I'd definitely be interested in the captures if you're happy to share them. They may also help us understand whether there is a reliable way to distinguish the protocol families without making assumptions based solely on the hardware class.

Thanks for pointing this out.

nord-:
The captures are public, so nothing to send privately. The A/B run that settled it is in [this comment on nord-/ha-ecovacs-mower#42](https://github.com/nord-/ha-ecovacs-mower/issues/42#issuecomment-5448858153): the reporter removed `77atlz` from our patched class list so the G1-800 ran on the stock V2 capabilities, and captured every `/q/` and `/p/` round trip. `clean_V2` was acked in about half a second and the mower resumed, `getCleanInfo_V2` was answered on the first attempt, against 900+ consecutive `errno 500` timeouts of plain `getCleanInfo` and unacknowledged `clean` on the patched build. Skim the earlier comments in that thread with care: two of them concluded "obeyed but not acknowledged" from runs that turned out to have been started from the official app, so the A/B at the end is the only evidence I would lean on.

On distinguishing the families up front: we looked for a way and did not find one. The class string is identical on both firmware branches. `fwVer` would decide it, but it only arrives in a command response header, and neither `GetDeviceList` nor `GetGlobalDeviceList` carries a version field, so it is not known at the point where the capabilities get frozen into `DeviceInfo.static`. That is what pushed us to decide from whether the mower answers rather than from any identity field.

The implementation is in [`deebot_patch/families.py`](https://github.com/nord-/ha-ecovacs-mower/blob/master/custom_components/ecovacs_mower/deebot_patch/families.py) and the `_AdaptiveFamily` mixin in [`deebot_patch/commands.py`](https://github.com/nord-/ha-ecovacs-mower/blob/master/custom_components/ecovacs_mower/deebot_patch/commands.py), merged in [nord-/ha-ecovacs-mower#70](https://github.com/nord-/ha-ecovacs-mower/pull/70). The shape, in case it is useful here: the selected family is sent first; on `errno 500` the other one is tried; the switch is committed per `did` only when the other family answers with `ret: ok`. Committing on the mere absence of `500` turned out to be wrong, since `errno 4200` (mower offline) and a plain timeout both satisfy that without the mower having said anything, and on 1.36.208 `getStats` fails intermittently with the same `errno 500`, so committing on the first failure would flip the dialect on every network blip. The state is keyed by `did` rather than held on the command object because `Capabilities.get_refresh_commands()` hands out shared instances while the clean action command is instantiated fresh on every call, so neither object is a stable home for it. The cost is one extra cloud timeout per genuine `500`, fine for a poll but noticeable when a user presses Start while the mower is briefly out of range.

One thing I take from your PR in return: `GoatCleanModeEvent` echoing `spotArea` back on pause/resume/stop. Our `CleanMower` sends `content: {type: auto}` on every action regardless of what started the job, and nobody has yet reported pausing an area job from Home Assistant on the classes we cover, so if the O1200 rejects a mismatched type that is something we would hit too. Good to know before someone does.

monsivar:
Thanks — this is excellent data, especially the clean A/B at the end of #42 and the details around when it is safe to commit a family switch.

The `errno 500` → try the other family → commit only on a positive `ret: ok` response distinction makes sense, as does keeping the decision per `did`. I also see why trying to determine it from device identity before capabilities are frozen does not appear viable.

For this PR I think keeping the change scoped to the physically verified O1200 (`2i0fns`) behaviour is still the safest option, rather than introducing the adaptive mechanism here. But your implementation looks like a very useful reference if support is generalized to GOAT classes where both firmware families exist.

And yes, the clean type on pause/resume/stop came directly from the O1200 captures. We observed the active `spotArea` type being carried through those actions, which is why `GoatCleanModeEvent` is retained rather than always falling back to `auto`.

I'll leave the PR scoped as it is for now while it awaits maintainer review, and address any design changes that come out of that review rather than expanding it pre-emptively.

Thanks for sharing all of this — it fills in a significant part of the `clean` vs `clean_V2` picture.

rhiermei:
Confirmation from a second GOAT O1200 LiDAR Pro (2i0fns) owner

Thanks for this PR – I can confirm the findings on my own O1200 LiDAR Pro, tested live today with the built-in Home Assistant Ecovacs integration (HA 2026.9.3).

Setup: GOAT O1200 LiDAR Pro, class 2i0fns, EU account (DE). Firmware: 1.13.10

1. clean_V2 is ignored completely. lawn_mower.start_mowing sends clean_V2 with {"act":"start","content":{"type":"auto"}}. The mower never answers (No response received for command "clean_V2") and stays docked. Every start triggered from HA since setup has failed this way (four occasions). Starting from the Ecovacs app with the same account works, so it is not an account-sharing permission issue.

2. clean works. Sending the identical payload with the command clean returns {"code":0,"msg":"ok"}, and the mower leaves the dock within about 3 seconds. Resume also works with clean and {"act":"resume","content":{"type":"auto"}}: docked with an open job → mowing after about 30 seconds.

3. Pitfall: an open job silently blocks start. If a job is still open (the app shows "End / Continue"), for example after lawn_mower.dock was called in the middle of a job, clean start still returns code 0 / ok, but the mower ignores it. Only resume gets it going again. It might be worth handling this in the integration (e.g. fall back to resume, or end the job when docking), because the "ok" response makes it look like a success.

4. Minor observations: getCleanInfo_V2 is not answered while the mower is docked. When sent home, the entity shows returning for only about 2 seconds and then paused until it is docked, while the app shows "docking" the whole time.

Workaround until this is merged: a small Python script using deebot-client inside the HA container (called via shell_command), with a Clean subclass that sends the payload above and skips the START/RESUME remapping in Clean._execute. Happy to share it if useful.

monsivar:
> Confirmation from a second GOAT O1200 LiDAR Pro (2i0fns) owner
> 
> Thanks for this PR – I can confirm the findings on my own O1200 LiDAR Pro, tested live today with the built-in Home Assistant Ecovacs integration (HA 2026.9.3).
> 
> Setup: GOAT O1200 LiDAR Pro, class 2i0fns, EU account (DE). Firmware: 1.13.10
> 
> 1. clean_V2 is ignored completely. lawn_mower.start_mowing sends clean_V2 with {"act":"start","content":{"type":"auto"}}. The mower never answers (No response received for command "clean_V2") and stays docked. Every start triggered from HA since setup has failed this way (four occasions). Starting from the Ecovacs app with the same account works, so it is not an account-sharing permission issue.
> 2. clean works. Sending the identical payload with the command clean returns {"code":0,"msg":"ok"}, and the mower leaves the dock within about 3 seconds. Resume also works with clean and {"act":"resume","content":{"type":"auto"}}: docked with an open job → mowing after about 30 seconds.
> 3. Pitfall: an open job silently blocks start. If a job is still open (the app shows "End / Continue"), for example after lawn_mower.dock was called in the middle of a job, clean start still returns code 0 / ok, but the mower ignores it. Only resume gets it going again. It might be worth handling this in the integration (e.g. fall back to resume, or end the job when docking), because the "ok" response makes it look like a success.
> 4. Minor observations: getCleanInfo_V2 is not answered while the mower is docked. When sent home, the entity shows returning for only about 2 seconds and then paused until it is docked, while the app shows "docking" the whole time.
> 
> Workaround until this is merged: a small Python script using deebot-client inside the HA container (called via shell_command), with a Clean subclass that sends the payload above and skips the START/RESUME remapping in Clean._execute. Happy to share it if useful.

Thanks for independently verifying this on another O1200 — especially the clean_V2 vs clean behavior on firmware 1.13.10. That gives us a useful second physical confirmation for the 2i0fns implementation in this PR.

The open-job behavior is also interesting. Since clean start can return ok while leaving an existing job untouched, I’d prefer not to infer a start→resume fallback in this PR without first establishing the relevant job-state signal and semantics. I’ll keep that as a follow-up point rather than expanding the command behavior here.

Thanks for testing this and documenting the results.

rhiermei:
Thanks for the quick reply. Fair enough on the open-job handling — keeping it as a follow-up point rather than widening this PR makes sense to me.

For anyone running into this before then: I work around it on the Home Assistant side rather than in the integration — a shell_command sends resume about a minute after start if the mower hasn't begun mowing by then. That covers the open-job case without guessing at job state inside the library.

Happy to test anything else on the O1200 (1.13.10) if a second device is useful at some point.

Issue #1762, opened 2026-08-14
T50 PRO OMNI (cuoipb): map SVG 500 because getMapInfo_V2 outlineVer=2 is rejected


hd00842:
### Checks

- [x] I have read the above text and I'm aware of the situation
- [x] I have searched the existing issues and no issue is describing my issue
- [x] I'm not creating an issue to add support for a new device/model.
- [x] I have installed the latest version

### The problem

The Ecovacs map image entity in Home Assistant never renders. `/api/image_proxy/image.*_map_*` returns HTTP 500 in ~6ms with body `500: Internal Server Error`. There is no aiohttp traceback because `EcovacsMap.image()` returns `None` when `Map.get_svg_map()` has no SVG.

This is not a missing hardware profile. `cuoipb` already aliases to the T50 map capabilities (`GetMapInfoV2`, `GetMapSetV2`, `GetMajorMap`, …). The robot also returns a large compressed `getMapInfo_V2` payload (~17KB). The library then drops it.

Root cause in `OnMapInfoV2._handle_body_data_dict` (`deebot_client/messages/json/map/__init__.py`):

- `outlineVer == "0"` is skipped (non-active map)
- only `outlineVer == "1"` is accepted
- any other value, including `"2"`, returns `HandlingResult.analyse()`

Debug log:

    Could not handle getMapInfo_V2 message: {..., 'outlineComplete': 0, 'outlineVer': '2', 'serial': '1', 'type': '0', 'using': 0}

After that:

- no `MapInfoEvent` is published, so Rust `MapInfo` stays empty
- `getMajorMap` returns `type=ol`, `value=""` (this model does not use the old PNG tile map)
- `generate_svg()` therefore returns `None`
- Home Assistant image proxy returns 500

Room names are partially available via `getMapSet_V2` (`Phòng bếp`, `Phòng khách`, plus two unnamed rooms), but room coordinates stay empty because they are expected from MapInfo V2.

This still happens on deebot-client 18.5.1 / HA 2026.8.1. The same `outlineVer != "1"` check is still on the current `dev` branch.

Expected: accept or parse `outlineVer: "2"` (or document that it is unsupported and do not 500 the image entity).

getCachedMapInfo: active map mid=1849463120, name="", using=1, built=1, angle=0

getMajorMap: {"cellHeight":1,"cellWidth":1,"mid":"1849463120","pieceHeight":800,"pieceWidth":800,"pixel":50,"type":"ol","value":""}

[...]

PR #1797, opened 2026-08-31
Accept outline version 2 in MapInfoV2 messages

andrewD-github:
## Problem

Some newer robots return `outlineVer: "2"` in `getMapInfo_V2` responses. The shared `OnMapInfoV2` handler currently rejects this before publishing `MapInfoEvent`, so the map stays empty and Home Assistant's image endpoint returns HTTP 500.

Refs #1762. Also reproduced on an X8 PRO OMNI (`n0vyif`, firmware `1.133.0`).

## Change

Accept outline versions `"1"` and `"2"` in the shared handler. Version `"0"` still skips inactive maps, and unknown versions are still rejected. This covers both requested responses and pushed map messages.

The observed V2 payload uses outline, room and block-line layers already supported by the existing Rust renderer. No renderer, hardware profile, dependency or Home Assistant changes are included.

## Validation

- Expanded version-handling tests for both message and command paths.
- Added a synthetic compressed map test that passes through each handler into the real SVG renderer, checking geometry and bounds without legacy map tiles. The fixture contains no real floor-plan data.
- Confirmed the four new V2 cases fail on unmodified `dev` at `5453eba`, then pass with the one-line change.
- Focused map suite: **62 passed, four existing SVG snapshots passed**. One existing pytest generator-parametrization deprecation warning remains.
- Ruff check/format and changed-file mypy (`--follow-imports=silent`) pass.
- Local native tests used the released 18.5.1 extension; the Rust source is unchanged between that tag and the tested upstream base. No fresh native build or full test suite was run locally.

On Home Assistant 2026.8.3 / deebot-client 18.5.1, the equivalent gate change was deployed via a temporary compatibility wrapper. The X8 map endpoint changed from HTTP 500 to HTTP 200, with room geometry, cleaning trace and position visible. A T9+ map continued to render. The captured X8 payload also rendered locally through the direct library patch.

Raw device logs, identifiers and the household's map are deliberately not attached. The reported T50/X11 devices in #1762 were not available for physical testing.

### DIFF
--- a/deebot_client/messages/json/map/__init__.py
+++ b/deebot_client/messages/json/map/__init__.py
@@ -84,7 +84,7 @@ class OnMapInfoV2(MessageBodyDataDict):
         if (outline_version := data.get("outlineVer")) == "0":
             # Skip it as it will be sent for non-active maps
             return HandlingResult.success()
-        if outline_version != "1":
+        if outline_version not in ("1", "2"):
             # Unsupported version
             return HandlingResult.analyse()
 
--- a/tests/commands/json/map/test_init.py
+++ b/tests/commands/json/map/test_init.py
@@ -596,7 +596,8 @@ async def test_getMapTrace() -> None:
     )
 
 
-async def test_getMapInfoV2() -> None:
+@pytest.mark.parametrize("outline_version", ["1", "2"])
+async def test_getMapInfoV2(outline_version: str) -> None:
     mid = "98100521"
     info = "KLUv/QRYmQAAW1siMSJdLFsiMiJdLFsiNiJdXbBRuA4="
     json, firmware_event = get_request_json(
@@ -609,7 +610,7 @@ async def test_getMapInfoV2() -> None:
                 "mid": mid,
                 "msgid": "",
                 "outlineComplete": 0,
-                "outlineVer": "1",
+                "outlineVer": outline_version,
                 "serial": "1",
                 "type": "0",
                 "using": 0,
@@ -635,7 +636,7 @@ async def test_getMapInfoV2_unsupported_version() -> None:
                 "mid": mid,
                 "msgid": "",
                 "outlineComplete": 0,
-                "outlineVer": "2",
+                "outlineVer": "3",
                 "serial": "1",
                 "type": "0",
                 "using": 0,
--- a/tests/messages/json/map/test_init.py
+++ b/tests/messages/json/map/test_init.py
@@ -1,18 +1,26 @@
 from __future__ import annotations
 
+from base64 import b64encode
+from compression.zstd import compress
 from typing import TYPE_CHECKING
 
+from defusedxml.ElementTree import fromstring
+import orjson
 import pytest
 
-from deebot_client.commands.json.map import GetMapSetV2
+from deebot_client.commands.json.map import GetMapInfoV2, GetMapSetV2
 from deebot_client.events import FirmwareEvent
 from deebot_client.events.map import MajorMapEvent, MapInfoEvent, MapSetType
 from deebot_client.message import HandlingState
 from deebot_client.messages.json import OnMapSetV2
 from deebot_client.messages.json.map import OnMajorMap, OnMapInfoV2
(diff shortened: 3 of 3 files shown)


raphaeldsc:
Same root cause, but a third outline version: my T50 PRO OMNI (`cuoipb`)
reports `outlineVer: "5"`, so this PR as written would not cover it.

- Home Assistant 2026.8.3, `deebot-client==18.5.1`
- Device class `cuoipb`, firmware `1.123.127`, hwVer `0.1.1`, wkVer `0.1.54`
- Symptom identical to #1762: `/api/image_proxy/image.*_map` returns HTTP 500
  with no traceback, because `EcovacsMap.image()` returns `None`.

Debug log (payload elided):

    DEBUG [deebot_client.message] Could not handle getMapInfo_V2 message:
    {'batid': 'aldebm', 'index': '1', 'info': 'KLUv/WCNSU1uAMqXdBARoCVp...',
     'infoSize': 19085, 'mid': '538622726', 'msgid': '', 'outlineComplete': 0,
     'outlineVer': '5', 'serial': '1', 'type': '0', 'using': 0}

All 10 `getMapInfo_V2` / `onMapInfo_V2` messages captured in one session
reported `outlineVer: "5"` — it is stable for this firmware, not intermittent.

**The payload is fully renderable by the current Rust code.** I extracted the
`info` field, base64-decoded and zstd-decompressed it: it expands cleanly to
exactly 19085 bytes, matching the advertised `infoSize`. It contains three
layers:

| Layer id in payload | `MapInfoType` in `src/map/map_info.rs` | Handler |
|---|---|---|
| `"1"` | `Outline` | `process_map_info_outline_entries` |
| `"2"` | `Room` | `process_map_info_room_entries` |
| `"6"` | `BlockLine` | `process_map_info_room_entries` |

All three are already implemented in 18.5.1. There is no `"5"` *layer* in the
data — note the numbering collision: the message field `outlineVer: "5"` is
unrelated to `MapInfoType::Unknown5`.

So for this device the only blocker is the version gate in
`OnMapInfoV2._handle_body_data_dict`. Would you consider accepting `"5"` in the
same allowlist as `"1"` and `"2"`?

More generally: since the gate protects a renderer that already ignores layer
types it does not recognise, rejecting the whole message on an unknown
`outlineVer` seems stricter than necessary. Accepting any version other than
`"0"` and letting the renderer skip unknown layers would stop this from
recurring with each new firmware — happy to defer to your judgement on that.

Caveat: I could not verify the render end-to-end, because the package lives in
the Home Assistant container and I could not patch the installed copy. The
evidence above is from the decompressed payload, not from a successful render.

reeesga:
Confirming this fixes #1762 on a T50, which I gather wasn't available for testing.

**Device:** DEEBOT T50 PRO OMNI — class `cuoipb`, fwVer `1.123.127`, hwVer `0.1.1`, wkVer `0.1.54`
**Stack:** Home Assistant 2026.9.0, deebot-client 18.5.1 (HAOS)

Applied only the gate change from this PR to the installed 18.5.1, nothing else:

```python
-        if outline_version != "1":
+        if outline_version not in ("1", "2"):
```

`/api/image_proxy/image.<vacuum>_map?token=…`:

- **Before:** `500`, 26 bytes, `text/plain` — no traceback, matching the symptom in #1762
- **After** (one core restart): `200`, `image/svg+xml`, 5805 bytes, valid SVG with per-room `<path class="r0">` geometry. Renders correctly in a Lovelace `picture` card with `image_entity`, showing all 7 rooms as distinct filled polygons.

Since `GetMapInfoV2(JsonCommandWithMessageHandling, OnMapInfoV2)` inherits the handler, the single change covers the command path as well — no separate fix needed there.

Caveat on scope: I only observed the map with the robot docked, so I haven't verified the live cleaning trace or position marker across a full run. Static map geometry and room layout are correct.

edenhaus:
I expect when the version is changing also the data structure is changing. Please analyse the new structure in detail and confirm all values are matching the old. I don't expect that one otherwise I would make no sense to update the version

PR #1804, opened 2026-09-06
Add support for DEEBOT X12 OmniCyclone (bmf9n5)

sekonicpr:
Adds a hardware definition for the DEEBOT X12 OmniCyclone (class bmf9n5) as a symlink to fd60kt (DEEBOT T50 OMNI)

class: 'bmf9n5'
deviceName: 'DEEBOT X12 OmniCyclone'
model: 'PLANCKS_DUST_BLACK_H_AMR'
materialNo: '110-2540-0301'
UILogicId: 'planckshcyc_ww_h_plancksh5'

Confirmed working on my unit (NA region, deebot-client 18.5.1) — device binds cleanly with the full entity set including map/position data, no getMapSet errors.

This is a separate variant from #1697 (c4wu9j) and #1809 (vihhjp) — same marketed model, different class code.


edenhaus (approved):

sekonicpr:
Pushed a follow-up converting the file to an actual symlink (mode 120000) — the original web-UI commit created it as a regular file, which is what caused the ruff/mypy failures. That commit is passing now.

Auto-merge got disabled by the new push — ready for another workflow approval whenever you have a moment. Thanks!

PR #1808, opened 2026-09-08
Add station water tank capability for T50 Max Pro Omni

djbill99-hue:
## Proposed change

Add an explicit station water-tank capability so integrations can distinguish docks with known water-tank fault support from auto-empty-only stations.

The new optional `CapabilityStationWaterTank` marker is exposed through `CapabilityStation.water_tank`. It is enabled for the shared T50 Max Pro Omni hardware definition used by `cuoipb`, where clean-water and dirty-water tank faults have been verified.

The existing `ErrorEvent` capability and refresh commands remain unchanged. This is a support marker only; it does not introduce a new event source or guarantee continuous tank-level reporting.

## Motivation

Home Assistant PR home-assistant/core#181667 adds actionable binary sensors for dock water-tank faults. The initial implementation exposed those sensors on devices that have generic error, station, and robot water capabilities but lack clean/dirty dock tanks.

An explicit capability allows Home Assistant to gate the sensors correctly without maintaining model allow-lists or inferring tank support from unrelated capabilities.

## Verification

* Verified on T50 Max Pro Omni hardware class `cuoipb`:

  * Error `322`: clean-water tank empty or not installed.
  * Error `323`: dirty-water tank full or not installed.
* Added tests confirming the marker is present for `cuoipb` and absent for `qhe2o2`, `yna5xi`, and `5xu9h3`.
* Confirmed `ErrorEvent` refresh commands remain `[GetError()]`.
* Full hardware test file: 11 passed.
* Pre-commit checks, including Ruff and mypy, passed.

## Related PR

* Home Assistant Core: https://github.com/home-assistant/core/pull/181667

### DIFF
--- a/deebot_client/capabilities.py
+++ b/deebot_client/capabilities.py
@@ -231,11 +231,6 @@ class CapabilitySettings:
     volume: CapabilitySet[VolumeEvent, [int]] | None = None
 
 
-@dataclass(frozen=True)
-class CapabilityStationWaterTank:
-    """Capabilities for station water tanks."""
-
-
 @dataclass(frozen=True, kw_only=True)
 class CapabilityStation:
     """Capabilities for the station."""
@@ -247,7 +242,7 @@ class CapabilityStation:
         auto_empty.Frequency,
     ]
     state: CapabilityEvent[StationEvent]
-    water_tank: CapabilityStationWaterTank | None = None
+    water_tank: CapabilityEvent[ErrorEvent] | None = None
 
 
 @dataclass(frozen=True, kw_only=True)
--- a/deebot_client/capabilities.py
+++ b/deebot_client/capabilities.py
@@ -231,6 +231,11 @@ class CapabilitySettings:
     volume: CapabilitySet[VolumeEvent, [int]] | None = None
 
 
+@dataclass(frozen=True)
+class CapabilityStationWaterTank:
+    """Capabilities for station water tanks."""
+
+
 @dataclass(frozen=True, kw_only=True)
 class CapabilityStation:
     """Capabilities for the station."""
@@ -242,6 +247,7 @@ class CapabilityStation:
         auto_empty.Frequency,
     ]
     state: CapabilityEvent[StationEvent]
+    water_tank: CapabilityStationWaterTank | None = None
 
 
 @dataclass(frozen=True, kw_only=True)
--- a/deebot_client/hardware/c8rj4y.py
+++ b/deebot_client/hardware/c8rj4y.py
@@ -18,7 +18,6 @@ from deebot_client.capabilities import (
     CapabilitySettings,
     CapabilitySetTypes,
     CapabilityStation,
-    CapabilityStationWaterTank,
     CapabilityStats,
     CapabilityWater,
     DeviceType,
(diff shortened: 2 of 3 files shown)


djbill99-hue:
@edenhaus When you have a chance, could you take a look at this? This capability is needed to correctly gate the water-tank sensors in home-assistant/core#181667; the current Core-side capability heuristic also matches auto-empty-only docks. The PR is intentionally small and keeps the existing ErrorEvent handling unchanged. Thanks!

deebot_client/capabilities.py
```diff
@@ -231,6 +231,11 @@ class CapabilitySettings:
     volume: CapabilitySet[VolumeEvent, [int]] | None = None
 
 
+@dataclass(frozen=True)
+class CapabilityStationWaterTank:
```
edenhaus: This is an empty class.. Looks strange
Are you sure that empty class is intentional?

How do we check for this capability?

djbill99-hue: Yes, the empty class was intentional as a support marker, with consumers checking station.water_tank is not None. I agree that an empty capability class is awkward, though. I’ll change this to an optional CapabilityEvent[ErrorEvent] on CapabilityStation and set it to CapabilityEvent(ErrorEvent, [GetError()]) for the verified T50 hardware. That keeps it consistent with the existing capability model, gives the field meaningful behavior, and consumers can still check whether station.water_tank is present.

nau-tic:
Thanks — an explicit marker is the right call, and I don't want to complicate it. I'm
flagging one device where the ErrorEvent 322/323 path can't fire, so it would be missed
by both this marker and the sensors in home-assistant/core#181667.

Device: DEEBOT X1 OMNI — class `1vxt52`, model `EINSTEIN_INT`, UILogicId `t10_ww_n_omni`.

What it reports (single session, 2026-09-13):

| Source                                   | Value while the clean tank was low |
|------------------------------------------|------------------------------------|
| onStationState -> content.error          | [301, 314]                         |
| getError (the ErrorEvent source)         | [0] — no error                     |

301 is already in ERROR_CODES as "FreshWaterBox empty". 314 is "module not installed",
which is expected on this tank-only model and is not a fault. The same content.error
field is present on the getStationState poll response.

Why the current approach misses it: this unit reports the clean-tank-low condition on
the station state, not through ErrorEvent. A gate on ErrorEvent codes 322/323 therefore
never fires for it, even though the fault is real. Marking `1vxt52` as water-tank-capable
alone would not fix that, because the HA sensor still reads ErrorEvent, which stays 0.

The library also drops the signal today: OnStationState (messages/json/station_state.py)
reads only state/type/motionState and discards content.error.

Proposal — model station-state errors as their own event, separate from ErrorEvent:

1. Add a StationErrorEvent carrying the code tuple, notified from the existing
   OnStationState handler. GetStationState subclasses OnStationState, so one event covers
   both the poll response and the push.
2. Leave ErrorEvent untouched. The HA tank sensors then accept either source: ErrorEvent
   322/323 for docks that use it, station-state 301/302 for docks like this one.

The alternative is to notify ErrorEvent from OnStationState for known tank codes. That
needs no HA change, but it mixes station state into the robot-error entity and changes
what the generic error sensor means — I'd rather not unless you prefer it.

Question: which would you prefer for docks that report tanks on the station state — a
separate station-state error event (my preference), or folding the codes into ErrorEvent?
I have the StationErrorEvent implementation ready and can open it as its own PR with
tests, following the split-PR convention.

(The dirty-tank-full code on this unit is expected to be 302; I haven't captured it live
yet.)

djbill99-hue:
Thanks, this is a great data point. I agree a separate StationErrorEvent is preferable to folding station-state errors into ErrorEvent.

This also shows that the current water_tank: CapabilityEvent[ErrorEvent] shape in this PR is too source-specific. If the separate station error event is accepted, I can adjust this capability so it represents water-tank support without assuming the underlying event source.

I’ll hold off on further changes here until the preferred event model is confirmed.

PR #1763, opened 2026-08-17
Add support for DEEBOT T50 OMNI Gen3 (re9vhz)

xlr23:
Adds support for the DEEBOT T50 OMNI Gen3 (device class `re9vhz`, model `FARADAYS_SE_WHITE_GEN3_INT`), by symlinking to the existing T50 OMNI (`fd60kt`) capabilities file — same pattern as #1649 which added support for the T50 PRO OMNI Gen3 (`npwtuz`).

Closes #1694

PR #1809, opened 2026-09-09
Add support for DEEBOT X12 OmniCyclone Care Complete (vihhjp)

dzukero:
The API reports the "Care Complete" bundle variant of the DEEBOT X12
OmniCyclone with hardware class `vihhjp`, for which no module exists,
so the device is discarded during discovery:

    deebot_client.api_client: Device class "vihhjp" not recognized.

The plain X12 OmniCyclone (class `c4wu9j`) is already supported and
aliased to `fd60kt` (DEEBOT T50 OMNI). Both `c4wu9j` and `vihhjp`
report the identical `UILogicId` ("planckshcyc_ww_h_plancksh5") and
`materialNo` prefix ("110-L003-04xx"), differing only in the EU
"Care" bundle SKU — the same signal check_for_similar_models.py uses
to group models. Alias `vihhjp` to the same `fd60kt` implementation
via the repository's symlink convention.

Follows the same pattern as recent similar PRs (e.g. #1755, #1763).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


edenhaus:
Did you verify that the vacuum is working? Will not add any AI generated stuff without proper testing as the history showed that assumptions of AI agents are sometimes completely wrong

dzukero:
> Did you verify that the vacuum is working? Will not add any AI generated stuff without proper testing as the history showed that assumptions of AI agents are sometimes completely wrong

Yes, and the assumptions were actually mine, I just had him do the PR :-)  I have two X12 OmniCyclones - one is c4wu9j (already supported) and this one - vihhjp which is identical with exception of the extra spare parts in the package.

edenhaus (approved):

